PR #2, opened 2016-02-19
test to ensure that local video unmute doesn't break large video

isymchych:


isymchych:
Yes, good idea. I think we should move as much js code as possible out of jitis-meet-torture :)

I'll create PR for lib-jitsi-meet and then will update this PR.

isymchych:
so I've added helper to the lib-jitsi-meet https://github.com/jitsi/lib-jitsi-meet/pull/19 and updated this PR to use the helper

paweldomas:
Hey Ivan, thanks for the PR! It is now merged, but there were few issues. Please see my commits following the merge for details. Thanks !

PR #3, opened 2016-02-19
add test to check if we can hang up the call

isymchych:


paweldomas:
Hi Ivan! The PR is now merged. I only renamed the test to EndConferenceTest as we have very similar class already and moved it to the end of the tests sequence. That's because the tests depend on each other and the next test was failing for me as the conference was closed for "owner" user. In general it would be best if each test would run independently, but we're not at this point yet, so please be aware of that for future commits. Thanks !

PR #5, opened 2016-02-24
Decreases disconnect timeout in conference migration.

paweldomas:
Jicofo with active health checks should detect failure faster.

To be merged only once https://github.com/jitsi/jicofo/pull/41 is merged and once we have page reload fixed in jitsi-meet.


gpolitis:
Taking it since I have to merge https://github.com/jitsi/jicofo/pull/41

PR #9, opened 2016-03-16
Implements test for max users limit in conference.

hristoterezov:


src/org/jitsi/meet/test/MaxUsersTests.java
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * @author Hristo Terezov
+ */
+public class MaxUsersTests
+    extends TestCase
+{
+    /**
+     * Number of participants in the call.
+     */
+    public static int MAX_USERS = 3;
```
paweldomas: It turns out that the plugin also counts jicofo as an user(which is fine). So when we use 3 it turns out that the max users limit is actually 2 and when you start the third one it will fail. I did not get that when reading the code. It would be more readable if some calculation would include 1 for jicofo and 1 for the user that will exceed the limit when constructing the length of "participants" array in the code below.

PR #11, opened 2016-03-21
Adjust "mute participant" test to work with JitsiPopover

isymchych:


src/org/jitsi/meet/test/MuteTest.java
```diff
@@ -157,14 +157,11 @@ public void ownerMutesParticipantAndCheck()
...
 
-        TestUtils.waitForDisplayedElementByXPath(
-            owner,
-            "//ul[@class='popupmenu']/li/a[@class='mutelink']",
-            5);
-
-        owner.findElement(
-                By.xpath("//ul[@class='popupmenu']/li/a[@class='mutelink']"))
-            .click();
+        TestUtils.waitMillis(1000);
```
gpolitis: Whenever I see these `waitMillis(...)` I'm wondering what they're trying to fix/prevent/mitigate :) Could you please add some comments explaining why we need this? I guess it's just the `waitForDisplayedElementByXPath` that's failing or not always working, could you please document this?

isymchych: that line was redundant, so I'll just remove it :)

damencho:
The popovers had been changed and this is no longer relevant.

PR #12, opened 2016-04-10
Implements performance time tests

hristoterezov:


bgrozev:
Can we use a different name for this test? I don't immediately understand what "performance time" means, and after reading the code I still don't understand. After reading jitsi-meet code, it seems that we only use the name "performance" because we use the "window.performance" API. I suggest using "ConnectionTimeTest".

hristoterezov:
@bgrozev Thanks for the name proposal! I've change it everywhere.

src/org/jitsi/meet/test/ConnectionTimeTest.java
```diff
@@ -0,0 +1,388 @@
...
+     * median of the subtracted values of the data from the passed time 
+     * measurement and the passed array. 
+     * @param previousStepTimes array with times that will be subtract
+     * @param s the time measurement
+     */
+    private void checkThreshold(Double[] previousStepTimes, TimeMeasurements s)
+    {
+        Double[] difference = (previousStepTimes == null)?  data[s.ordinal()] :
+            subtractArrays(previousStepTimes, data[s.ordinal()]);
+        assertTrue(getMedian(difference) < s.getThreshold());
```
damencho: We need to add here some meaningful message what had actually failed, if it fail.

PR #13, opened 2016-04-14
Adds checks for the data channel being opened.

bgrozev:


src/org/jitsi/meet/test/ConnectionTimeTest.java
```diff
@@ -91,8 +91,18 @@
...
+            ICE_CONNECTED, 200.0),
+
+        // The data channel should open about 2 RTTs after DTLS completes, so
+        // this threshold should go down to something like 200ms.
+        // However, there is currently a bug in jitsi-videobridge which adds a
+        // delay of about 3 seconds. Another bug, recently fixed, would cause
+        // a delay of ~15 seconds, which is why we now use a threshold of 4s.
+        DATA_CHANNEL_OPENED(
+            "return APP.conference._room.getConnectionTimes()"
+                + "['data.channel.opened']",
```
paweldomas: Isn't the key different here:
https://github.com/jitsi/lib-jitsi-meet/pull/72/files#diff-40e35c919311b508dc5b45a7bf028f54R1076
?

bgrozev: Yeah, Hristo just pinged me and I fixed it in the lib. Thanks!

PR #18, opened 2016-05-23
Added integration test to test jitsi-meet behavior when list of devices changes

tsareg:
This integration test covers only couple of base cases. Another cases might be too time consuming and also rely much more on mocking JS code, so I see no sense implementing them.


tsareg:
This should be merged after https://github.com/jitsi/lib-jitsi-meet/pull/119 is merged into master

tsareg:
PR is no longer valid, will re-create

PR #19, opened 2016-06-01
Added automated tests for "Follow Me" feature

tsareg:


src/org/jitsi/meet/test/FollowMeTest.java
```diff
@@ -0,0 +1,194 @@
...
+    {
+        System.err.println("Start testFollowMeCheckboxVisibleOnlyForModerator");
+
+        WebDriver owner = ConferenceFixture.getOwner();
+        WebDriver secondParticipant = ConferenceFixture.getSecondParticipant();
+
+        checkFollowMeCheckboxVisible(owner, true);
+        checkFollowMeCheckboxVisible(secondParticipant, false);
+
+        owner.findElement(By.id("followMeCheckBox")).click();
```
damencho: This enables follow me, but never disables it. Currently when running the tests some of the tests after that fail for me. Generally the tests should leave participants in the state they have found it when starting. Could you please check this one, does all tests pass for you running them against beta?

src/org/jitsi/meet/test/FollowMeTest.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * Checks if launching and then exiting Etherpad is executed for the second
+     * participant.
+     */
+    public void testShareDocumentCommandsAreFollowed() {
```
damencho: This test will fail if no etherpad is configured, but we still want follow me to be tested. You can check: https://github.com/jitsi/jitsi-meet-torture/blob/master/src/org/jitsi/meet/test/EtherpadTest.java#L84
Currently pr-testing has no etherpad configured and pushing this without a check that skips the test will make all pr-testing to fail. Could you please add this check?

damencho:
The commits were cherry-picked in a torture branch and with some fixes everything is now merged.

PR #28, opened 2016-07-11
Changes clickOnLocalVideo util to click on localVideoContainer instead of the focusIndicator

hristoterezov:


lyubomir:
This PR looks OK to me from the point of view of reading & not testing i.e. like https://github.com/jitsi/lib-jitsi-meet/pull/166 & https://github.com/jitsi/jitsi-meet/pull/724.

PR #33, opened 2016-09-07
BugFix - StartMutedTest:

FearTheBadger:
- changed to owner for the click of 'toolbar_button_mute'
- made the secondParticipant the observer and the owner the testee to match the test setup.

### DIFF
--- a/src/org/jitsi/meet/test/StartMutedTest.java
+++ b/src/org/jitsi/meet/test/StartMutedTest.java
@@ -124,12 +124,12 @@ public class StartMutedTest
 
         // Unmute and see if the audio works
         MeetUIUtils.clickOnToolbarButton(
-            secondParticipant, "toolbar_button_mute");
+            owner, "toolbar_button_mute");
         ((JavascriptExecutor) owner)
             .executeScript(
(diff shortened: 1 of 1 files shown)


damencho:
I don't think this is relevant. It does the same thing current test do. 
It doesn't matter if you unmute owner and check on second participant side whether it is unmuted or if you unmute second participant and check on owner side whether it is unmuted.
What do you try ot fix? We have this test running without a failure for some time.

FearTheBadger:
Interesting, it fails for us without this change.  I might have the logic incorrect.

Reading through the current master logic:
- Starts the owner muted.
- Clicks the second participants mute button (which mutes that second participant, as they didn't start muted).
- Checks to see that the second participant is un-muted. (This is where we fail as they were just muted.)

damencho:
Nope, this is not correct, when the owner opens the room with config.startAudioMuted=1&config.startVideoMuted=1, this enables startMuted for the conference, the same as if it goes and clicks it in the settings. So when second participant joins, he is muted.

FearTheBadger:
With this information, I'll kill this PR.  I will have to track down why we are failing the test, and why our second participant is not starting muted.

Thanks.

PR #40, opened 2016-09-19
Fixes audio mute, video mute and display name tests.

yanas:


src/org/jitsi/meet/test/util/TestUtils.java
```diff
@@ -157,10 +157,8 @@ public static void waitForElementByXPath(
         long timeout)
     {
         new WebDriverWait(participant, timeout)
-            .until(new ExpectedCondition<Boolean>()
-            {
-                public Boolean apply(WebDriver d)
-                {
+            .until(new ExpectedCondition<Boolean>() {
```
paweldomas: { should probably be in the next line. The same applies for few other changes below

PR #53, opened 2016-10-19
After clicking remote or local video give time for the video/avatar to change.

damencho:
…o change.

### DIFF
--- a/src/org/jitsi/meet/test/util/MeetUIUtils.java
+++ b/src/org/jitsi/meet/test/util/MeetUIUtils.java
@@ -554,6 +554,25 @@ public class MeetUIUtils
     }
 
     /**
+     * Checks whether the large video has changed to the desired resource.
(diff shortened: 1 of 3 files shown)


src/org/jitsi/meet/test/util/MeetUIUtils.java
```diff
@@ -554,6 +554,25 @@ public static String getLargeVideoID(WebDriver driver)
     }
 
     /**
+     * Checks whether the large video has changed to the desired resource.
+     * @param driver the driver to check
+     * @param resource the expected resource
+     */
+    public static void assertWrongAvatarOnLarge(
```
paweldomas: "assertWrong" feels a bit misleading, like this assertion will check if the avatar actually *is wrong*/different than the tested one. But int the check we're looking if it's *right*(asserting equality).

PR #57, opened 2016-11-03
Add new method to TestUtils; fix the test

BeatC:
Fix the test from https://github.com/jitsi/jitsi-meet/pull/1107

### DIFF
--- a/src/org/jitsi/meet/test/util/TestUtils.java
+++ b/src/org/jitsi/meet/test/util/TestUtils.java
@@ -305,6 +305,14 @@ public class TestUtils
                 });
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


src/org/jitsi/meet/test/util/TestUtils.java
```diff
@@ -305,6 +305,25 @@ public Boolean apply(WebDriver d)
                 });
     }
 
+    public static void waitForElementContainsClassByXPath(
```
yanas: LGTM, we just need some javadoc above. This isn't a blocker for merging the commit, but please add it in a separate PR rather sooner than later, so that we don't forget.

PR #67, opened 2017-02-10
feat(ss): Add test for size

hristoterezov:


damencho:
Seems there is some problem, not sure is it the test or the code itself.

org jitsi meet test desktopsharingimitationtest testdesktopsharing-owner
org jitsi meet test desktopsharingimitationtest testdesktopsharing-participant

hristoterezov:
Actually the second screenshot looks fine because we are trying to keep ratio of the video and that's why there is a space below the video.

PR #80, opened 2017-05-08
feat(audio-only): Basic integration tests

virtuacoplenny:
These tests have the owner enable audio only, then verify video is
not display, then disable audio only, and finally verify video is
displayed.


virtuacoplenny:
These tests need to be redone because the audio only button is being removed from the toolbar.

PR #82, opened 2017-05-23
feat(audio-only): Basic integration tests

virtuacoplenny:
These tests have the owner enable audio only, then verify video is
not display, then disable audio only, and finally verify video is
displayed.


virtuacoplenny:
I've rebased onto the latest jitsi-meet-torture master and then tested against beta and the new tests still pass.

PR #86, opened 2017-06-19
Protect against exceptions when checking if an element is displayed

virtuacoplenny:


gpolitis:
Not strictly required, but it would be nice to have a little comment as to why the try catch is necessary.

virtuacoplenny:
I've added a comment in the catch block explaining why the catch can happen.

PR #90, opened 2017-07-12
ref(small-video): video element xpath as descendant not child

virtuacoplenny:
Instead of assuming the thumbnail's video element is a direct
child of another element, perform a descendant search. This will
help keep the tests flexible while reactification work is ongoing.


virtuacoplenny:
Remade with a different branch name here: https://github.com/jitsi/jitsi-meet-torture/pull/92

PR #93, opened 2017-07-19
Do not assume remote video menu dom structure and change to click

virtuacoplenny:


src/test/java/org/jitsi/meet/test/MuteTest.java
```diff
@@ -159,29 +159,25 @@ public void ownerMutesParticipantAndCheck()
 
         WebElement elem = owner.findElement(By.xpath(
             TestUtils.getXPathStringForClassName("//span", "remotevideomenu")
-            + "/i[@class='icon-thumb-menu']"));
+            + "//i[@class='icon-thumb-menu']"));
```
bbaldino: is it worth having constants for these selectors?

virtuacoplenny: Yes, it's worthwhile but almost all these files have these selectors inlined. What needs to happen, and what I'd really like to do this year, is start moving these over into page objects to encapsulate these lookups, moving these to constants in the process.

virtuacoplenny: (In other words, I agree but not sure I should do it at this point for consistency.)

virtuacoplenny: But then again why not? It is better for readability. I'll just make the change now.

PR #96, opened 2017-09-07
fix(1-on-1): adjust toolbar hide times, change hover selector

virtuacoplenny:
- The toolbar display durations have been reduced to compensate
  for a torture test machine that might be a little slow.
- The selector for the mousehover target, for showing remote
  videos, has been changed to one that is guaranteed to be
  available.


damencho:
The nightly tests had failed this test last 5 or 6 runs, I've tested this change and it succeeded several times in a row. Merging it.

PR #99, opened 2017-10-24
fixes/enhancements to the psnr test

bbaldino:
SQUASH BEFORE MERGING


scripts/psnr-test.sh
```diff
@@ -16,11 +16,20 @@
...
-INPUT_FRAME_DIR=$2
-RESIZED_FRAME_DIR=$3
+OUTPUT_FRAME="$1"
+INPUT_FRAME_DIR="$2"
+RESIZED_FRAME_DIR="$3"
+QR_CODE_DECODER="zbarimg"
 
-FRAME_NUMBER=$(java -jar $SCRIPT_DIR/javase-3.2.2-SNAPSHOT-jar-with-dependencies.jar $OUTPUT_FRAME |head -3|tail -1)
+
+PATH_TO_QR_DECODER=$(which $QR_CODE_DECODER)
```
gpolitis: would it make sense to leave the old decoder in a comment for future reference?

scripts/push_psnr_results.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/python
```
gpolitis: Adding the licence header and a brief description of what this does would be nice

scripts/psnr-test.sh
bbaldino: i figured the git history would be enough (i did still leave the jar in the repo).  do you think i should add the old one back and leave it commented?

scripts/push_psnr_results.py
bbaldino: good idea, will do

scripts/psnr-test.sh
gpolitis: Actually feel free to disregard my comment. I suppose you have tested with zbarimg and you got no false negatives thus I don't see why we would ever want to go back to ZXing.

gpolitis:
LGTM. I'll wait a couple of hours in case anybody else wants to comment on it and then I'll merge it

src/test/java/org/jitsi/meet/test/PSNRTest.java
```diff
@@ -196,6 +207,20 @@ public void testPSNR()
                 File outputFrameFile = new File(outputFrame);
                 outputFrameFile.delete();
             }
+            float averagePsnr = totalPsnr / framesCount;
```
bgrozev: Would we also want to fail the test if the average is below a certain threshold (the per-frame threshold is very very low)?

scripts/psnr-test.sh
```diff
@@ -16,11 +16,20 @@
...
 SCRIPT_DIR=$(dirname $0)
-OUTPUT_FRAME=$1
-INPUT_FRAME_DIR=$2
-RESIZED_FRAME_DIR=$3
+OUTPUT_FRAME="$1"
+INPUT_FRAME_DIR="$2"
+RESIZED_FRAME_DIR="$3"
+QR_CODE_DECODER="zbarimg"
 
-FRAME_NUMBER=$(java -jar $SCRIPT_DIR/javase-3.2.2-SNAPSHOT-jar-with-dependencies.jar $OUTPUT_FRAME |head -3|tail -1)
```
bgrozev: Do we still need that jar if we're using zbarimg instead?

src/test/java/org/jitsi/meet/test/PSNRTest.java
bbaldino: we could, but i think the per-frame one will be even more sensitive, no? (unless you mean having the average required to be an even higher value?)

scripts/psnr-test.sh
bbaldino: no...and i think you're right, we should remove it.  i'll delete it.

bbaldino:
@gpolitis all good here?

PR #106, opened 2017-11-28
feat: basic mobile test

paweldomas:
This adds a mobile test which will:
 1. Go to the welcome page
 2. Enter room name
 3. Try to join
 4. Once, joined - show the toolbar and click hangup.


src/main/java/org.jitsi.meet.test/mobile/ToolbarView.java
```diff
@@ -0,0 +1,56 @@
...
+import io.appium.java_client.pagefactory.*;
+
+import org.jitsi.meet.test.mobile.base.*;
+
+import org.openqa.selenium.*;
+
+/**
+ * Page object for the mobile toolbar area.
+ */
+public class ToolbarView extends AbstractMobilePage
```
saghul: Don't we call it toolbox? Also, how about getting a reference to the toolbox, and then geting the hangup button out of it? That would introduce the machinery we need to click other buttons in the future. WDYT?

src/main/java/org.jitsi.meet.test/mobile/WelcomePageView.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Page object for the mobile welcome page.
+ */
+public class WelcomePageView extends AbstractMobilePage
+{
+    /**
+     * Conference room name input locator.
+     */
+    @AndroidFindBy(accessibility = "Input room name.")
+    @iOSFindBy(accessibility = "Input room name.")
```
saghul: Since these 2 will almost always be the same, is there a way to combine them into one? Something like `@MobileFindBy` perhaps.

src/test/java/org/jitsi/meet/test/mobile/base/AbstractBaseTest.java
```diff
@@ -41,15 +43,63 @@
...
      * Current mobile driver instance.
      */
     private static AppiumDriver<WebElement> driver;
 
     /**
-     * The current room name used.
+     * The OS currently being tested ('android' or 'ios').
      */
-    public static String currentRoomName;
+    private static String os;
```
saghul: platform? since 'web' can also be a valid one, I guess?

src/main/java/org.jitsi.meet.test/mobile/WelcomePageView.java
paweldomas: I was trying to figure this out, but there's no way currently to do that as far as I know. I've tried asking the community, but no response so far:

https://discuss.appium.io/t/find-by-accessibility-id-on-both-android-and-ios-using-one-annotation/19746

src/main/java/org.jitsi.meet.test/mobile/ToolbarView.java
paweldomas: You mean a reference to the toolbox container element ? I'm not sure if that makes sense with how the things are currently being located...

paweldomas: And I can rename it to toolbox - sure

src/test/java/org/jitsi/meet/test/mobile/base/AbstractBaseTest.java
paweldomas: I next PR this will go away, because it will be possible to run Android, iOS and web all at the same time. There will be a separate class called MobileParticipant that will be dealing with that.

saghul: Ack.

src/main/java/org.jitsi.meet.test/mobile/WelcomePageView.java
saghul: Haha, great minds think alike, I guess :-P Maybe we can at least make the string a static constant to avoid typos?

src/main/java/org.jitsi.meet.test/mobile/ToolbarView.java
saghul: Yes, I meant the container. Then we'd find the buttons inside the container.

paweldomas:
About the docs: I wanted to go either with you or Lyubo through the process step by step to write the docs, because I don't remember anymore what is needed exactly - and there were few details.

src/main/java/org.jitsi.meet.test/mobile/ToolbarView.java
paweldomas: It will not be easier this way. You'll have to add accessibility ID to the container and then each button will still need to have something that identifies it, so that we can find it - I guess that would be accessibility Id, so...

saghul: That sounds good to me though :-) But as I saiid, I think you are better qualified to make a judgement here.

PR #110, opened 2018-01-04
Moves tests to testng

damencho:


src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
```diff
@@ -0,0 +1,195 @@
...
+    {
+        Assert.assertEquals(obj1, obj2, message);
+    }
+
+    /**
+     * Wraps assertFalse.
+     * @param message the message to show.
+     * @param value the value to check.
+     */
+    public void assertFalse(String message, boolean value)
```
paweldomas: As we've discussed this offline, the purpose of having these methods is to have the other tests ported more easily - with less changes, but seeing those methods here feels weird. There should be at least a comment with an explanation and a plan to be removed as next step (I'm reading commits one by one and don't know yet if this was changed in next commits)

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
```diff
@@ -0,0 +1,177 @@
...
+     * and stops the third participant, if it is not stopped.
+     * participants(owner and 'second participant').
+     */
+    public void ensureTwoParticipants()
+    {
+        waitForParticipant1ToJoinMUC();
+        waitForSecondParticipantToConnect();
+
+        // TODO do we need this???
+        //closeThirdParticipant();
```
paweldomas: If there are 3 participants then it should be closed. That's what the method name says - that it will make sure there are exactly 2 participants.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
```diff
@@ -0,0 +1,177 @@
...
+                    .createParticipant("web.participant1");
+        }
+
+        if (participant1.isHungUp())
+        {
+            // join room
+            participant1.joinConference(currentRoomName);
+        }
+
+        MeetUtils.waitForParticipantToJoinMUC(participant1.getDriver(), 15);
```
paweldomas: Ideally "waitForParticipantToJoinMuc" would be a participant's method.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
```diff
@@ -0,0 +1,177 @@
...
+
+        // TODO do we need this???
+        //closeThirdParticipant();
+    }
+
+    /**
+     * Waits until the owner joins the room, creating and starting the owner
+     * if it hasn't been started.
+     */
+    public void waitForParticipant1ToJoinMUC()
```
paweldomas: I don't like this method and even more the call to "isHangUp()" inside of it. This procedure should be encapsulated inside participant.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
```diff
@@ -0,0 +1,177 @@
...
+    {
+        if (participant1 == null)
+        {
+            participant1
```
paweldomas: Ideally there would be a method "get1stParticipant()" which would check if it's null and create it if needed. I bet there could be more places that need similar logic.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
```diff
@@ -0,0 +1,177 @@
...
+            participant2 = ParticipantFactory.getInstance()
+                .createParticipant("web.participant2");
+        }
+
+        if (participant2.isHungUp())
+        {
+            participant2.joinConference(currentRoomName);
+        }
+
+        MeetUtils.waitForParticipantToJoinMUC(participant2.getDriver(), 10);
```
paweldomas: All those things should be part of participant.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Starts the owner, if it isn't started and hangups all other participants.
+     */
+    public void closeAllParticipantsExceptTheOwner()
+    {
+        waitForParticipant1ToJoinMUC();
+
+        if (participant2 != null && !participant2.isHungUp())
```
paweldomas: isHangUp() should be called automagically inside hangUp() without the need for doing it outside of the class.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
```diff
@@ -0,0 +1,177 @@
...
+        {
+            participant2.hangUp();
+        }
+
+        if (participant3 != null && !participant3.isHungUp())
```
paweldomas: This conditions ask for a method "hangupXParticipant". It feels like it's a logic that would be needed in more places, but I haven't check that yet.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * Waits until {@code thirdParticipant} has joined the conference (its ICE
+     * connection has completed and has it has sent and received data).
+     */
+    public void waitForThirdParticipantToConnect()
+    {
+        if (participant3 == null)
+        {
+            participant3 = ParticipantFactory.getInstance()
```
paweldomas: getThirdParticipant()

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
```diff
@@ -0,0 +1,177 @@
...
+                .createParticipant("web.participant3");
+        }
+
+        if (participant3.isHungUp())
```
paweldomas: It reads weird or is hard to understand. If participant "is hung up" then join, but otherwise there's no need to call joinConference like it would be doing that automatically ? What if it was not called previously ? For me it would read easier if "joinConference" would be doing the "isHungUp" check.

src/test/java/org/jitsi/meet/test/base/Participant.java
```diff
@@ -0,0 +1,214 @@
...
+ */
+package org.jitsi.meet.test.base;
+
+import org.jitsi.meet.test.util.*;
+import org.openqa.selenium.*;
+
+import java.util.concurrent.*;
+
+/**
+ * The participant instance holding the {@link WebDriver}.
```
paweldomas: It seems that many of the things here are web specific. Why isn't this class called WebParticipant ?

src/test/java/org/jitsi/meet/test/base/FailureListener.java
```diff
@@ -0,0 +1,362 @@
...
+     *
+     * @param fileName the filename to use.
+     */
+    private void takeScreenshots(
+        String fileName, AbstractBaseTest testInstance)
+    {
+        takeScreenshot(testInstance.participant1.getDriver(),
+            fileName + "-owner.png");
+
+        if(testInstance.participant2 != null)
```
paweldomas: no space after if - few more occurrences in this commit, probably brackets would make it easier to read since those statements are multiline

src/test/java/org/jitsi/meet/test/base/FailureListener.java
```diff
@@ -0,0 +1,362 @@
...
+                + fileName);
+            return;
+        }
+
+        File scrFile = takesScreenshot.getScreenshotAs(OutputType.FILE);
+        File destFile = new File(outputScreenshotsParentFolder, fileName);
+        try
+        {
+            FileUtils.copyFile(scrFile, destFile);
+        } catch (IOException ioe)
```
paweldomas: break line before catch

src/test/java/org/jitsi/meet/test/base/FailureListener.java
```diff
@@ -0,0 +1,362 @@
...
+            if(log == null)
+                return;
+
+            FileUtils.write(
+                new File(outputLogsParentFolder, fileName),
+                (String)log);
+        }
+        catch (Exception e)
+        {
+            //e.printStackTrace();
```
paweldomas: why is that exception swallowed ?

src/test/java/org/jitsi/meet/test/base/FailureListener.java
```diff
@@ -0,0 +1,362 @@
...
+        try
+        {
+            BufferedWriter out = new BufferedWriter(new FileWriter(
+                new File(outputLogsParentFolder,
+                    fileNamePrefix + ".tdump")));
+            out.write(dump.toString());
+            out.flush();
+            out.close();
+        }
+        catch (IOException e){}
```
paweldomas: ignored exception, you can also use new java 8 syntax in all file operations:
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/test/java/org/jitsi/meet/test/base/FailureListener.java
```diff
@@ -0,0 +1,362 @@
...
+                    new File(outputLogsParentFolder,
+                        "chrome" + suffix + ".log"),
+                    new File(outputLogsParentFolder,
+                        fileNamePrefix + suffix + "-chrome" + extension)
+                );
+            }
+        }
+        catch (IOException e)
+        {
+            // cannot create file or something
```
paweldomas: haha

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
```diff
@@ -60,30 +89,54 @@ public void ensureTwoParticipants()
...
     {
         if (participant1 == null)
         {
             participant1
                 = ParticipantFactory.getInstance()
-                    .createParticipant("web.participant1");
+                .createParticipant("web.participant1");
         }
 
         if (participant1.isHungUp())
```
paweldomas: calling " startParticipant1" method will have no effect if participant is "not hung up" ? it feels weird, if it's expected to be called in some particular state then it should be checked if the assertion is right and if not throw IllegalStateException.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
```diff
@@ -133,22 +201,39 @@ public void waitForSecondParticipantToConnect()
...
 
     /**
      * Starts the owner, if it isn't started and hangups all other participants.
      */
-    public void closeAllParticipantsExceptTheOwner()
+    public void hangUpAllParticipantsExceptTheOwner()
+    {
+        waitForParticipant1ToJoinMUC(null, null);
+
+        if (participant2 != null && !participant2.isHungUp())
```
paweldomas: it asks for "hangupParticipant2" method, see how this logic is duplicated in many other places...

src/test/java/org/jitsi/meet/test/util/MeetUtils.java
```diff
@@ -224,8 +224,9 @@ public Boolean apply(WebDriver driver)
         }
         catch(Throwable error)
```
paweldomas: I don't get why Throwable is caught here given that there is specific timeout exception type. It could be interesting if some other error is thrown here - it could be logged at least.

src/test/java/org/jitsi/meet/test/PeerConnectionStatusTest.java
```diff
@@ -239,21 +209,21 @@ public void test2ndPeerInterruptedDuringCall()
...
 
         // 4. Select remote again
         MeetUIUtils.selectRemoteVideo(owner, secondPeer);
 
         // Mute video(this is not real live scenario, but we want to make
         // sure that the grey avatar is displayed if the connection gets
         // disrupted while video muted)
-        new StopVideoTest("stopVideoOnParticipantAndCheck")
-            .stopVideoOnParticipantAndCheck();
+        //        new StopVideoTest("stopVideoOnParticipantAndCheck")
```
paweldomas: why are those commented out (without any comment )? Is it for the work in progress ?

damencho: I will check it, maybe I missed it while working on it.

src/test/java/org/jitsi/meet/test/DisplayNameTest.java
```diff
@@ -28,23 +29,49 @@
...
+    }
+
+    /**
+     * Default constructor.
+     */
+    public DisplayNameTest()
+    {}
+
+    /**
+     * Constructs DisplayNameTest with already allocated participants.
```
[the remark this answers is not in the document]
damencho: RingOverlayTest and ReloadTest, we have similar stuff about StopVideoTest and SwitchVideoTest ... maybe more.

src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
damencho: Updates, there is a commit porting all assert calls to testng.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
damencho: Updated.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
damencho: Waits are now moved to Participant.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
damencho: Removed.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
damencho: Moved.

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
damencho: Fixed

src/test/java/org/jitsi/meet/test/base/Participant.java
damencho: Introduced WebParticipant.

src/test/java/org/jitsi/meet/test/base/FailureListener.java
damencho: There is a commit fixing all formatting that was introduced earlier.

src/test/java/org/jitsi/meet/test/base/FailureListener.java
damencho: Fixed

src/test/java/org/jitsi/meet/test/base/FailureListener.java
damencho: Fixed

src/test/java/org/jitsi/meet/test/base/AbstractParticipantHelper.java
damencho: Fixed

src/test/java/org/jitsi/meet/test/RestTests.java
```diff
@@ -90,7 +92,9 @@ private void checkConferencesThroughREST()
 
         JsonArray confs = new Gson().fromJson(conferences, JsonArray.class);
         if(confs == null || confs.size() == 0)
-            assertFalse("Expected at least one conference", true);
```
paweldomas: haha I love this old code: assertFalse(..., true)

src/test/java/org/jitsi/meet/test/SharedVideoTest.java
```diff
@@ -339,10 +344,12 @@ public void backToSharedVideoTest()
             .click();
         TestUtils.waitMillis(1000);
 
-        assertEquals("Video not displayed:", true,
+        assertEquals(
```
paweldomas: why not assertTrue (the same if few other places) ?

src/test/java/org/jitsi/meet/test/SharedVideoTest.java
```diff
@@ -367,12 +374,13 @@ public void stopSharingTest()
 
         // video should be visible on both sides
         WebDriver secondParticipant = participant2.getDriver();
-        assertEquals("Video not displayed:", true,
-            owner.findElement(By.id("sharedVideoIFrame"))
-                .isDisplayed());
-        assertEquals("Video not displayed:", true,
-            secondParticipant.findElement(By.id("sharedVideoIFrame"))
-                .isDisplayed());
+        assertEquals(
```
paweldomas: assertTrue

src/test/java/org/jitsi/meet/test/SharedVideoTest.java
```diff
@@ -367,12 +374,13 @@ public void stopSharingTest()
...
+            true, owner.findElement(By.id("sharedVideoIFrame")).isDisplayed(),
+            "Video not displayed:");
+        assertEquals(
```
paweldomas: assertTrue

src/test/java/org/jitsi/meet/test/SharedVideoTest.java
```diff
@@ -503,9 +511,10 @@ private void shareVideoBeforeOthersJoin(boolean pause)
 
         checkPlayerLoadedAndInState(secondParticipant, stateToCheck);
 
-        assertEquals("Video not displayed:", true,
-            secondParticipant.findElement(By.id("sharedVideoIFrame"))
-                .isDisplayed());
+        assertEquals(
```
paweldomas: assertTrue

src/test/java/org/jitsi/meet/test/ConnectionTimeTest.java
```diff
@@ -346,6 +346,9 @@ private void failIfContainsNull(Double[] array)
         {
             refreshSecondParticipant();
 
+            // just ping the first participant, to avoid timeout of the session
```
paweldomas: why don't we increase the timeout ? I remember the need to adjust them for hybrid anyway

src/test/java/org/jitsi/meet/test/base/Participant.java
```diff
@@ -86,6 +87,13 @@ public void joinConference(String roomName)
      */
     public void joinConference(String roomName, String fragment)
     {
+        // not hanguped, so not joining
+        if (!this.hungUp)
+        {
+            TestUtils.print("Not joining " + roomName + ", already joined.");
```
paweldomas: how do you know that it's the given roomName that is joined ? it could be anything

src/test/java/org/jitsi/meet/test/util/MeetUtils.java
```diff
@@ -297,56 +297,56 @@ public static boolean isXmppConnected(WebDriver participant)
...
-     *
-     * @param participant the participant.
-     * @param timeout timeout in seconds.
-     */
-    public static void waitForIceConnected(WebDriver participant, long timeout)
-    {
-        TestUtils.waitForBoolean(
-            participant, ICE_CONNECTED_CHECK_SCRIPT, timeout);
-    }
+//    /**
```
paweldomas: there's a bunch of commented out code in all places, was this removed in next commits ?

src/test/java/org/jitsi/meet/test/ConnectionTimeTest.java
damencho: So this is something else. This timeout is not specific to the test but to the grid. So I was testing and initially, this timeout was very high and if a test crashes the node stay unusable for a long time, so I changed the timeout to 30 seconds. Which means that if a session to a selenium grid node does not receive anything in 30 seconds that session will be closed and the node will be freed. 
In case of this test, we open the first participant and then start reloading the second one for more than a minute (10 reloads). And without touching the first session the timeout is hit and the session expires on the grid. Just executing something fixes the problem, without relying on the value that is on the grid.

src/test/java/org/jitsi/meet/test/util/MeetUtils.java
damencho: Yep, there is a commit cleaning those.

src/test/java/org/jitsi/meet/test/base/Participant.java
damencho: Yep, will fix that.

src/test/java/org/jitsi/meet/test/ConnectionTimeTest.java
paweldomas: I know what this is about. The same is with mobile/hybrid when waiting for one mobile to come up other will timeout. There must be a cleaner way, keep alive or something

paweldomas: and the case you've described what is it that crashes ? java test process ?

paweldomas: one of first google results: http://developers.perfectomobile.com/display/TT/Keeping+long+Selenium+scripts+alive
If we'd have a keep alive thread bound to every participant which dies with java process I guess it would be fine. But given it's a bigger effort I would be fine leaving it for another step with a FIXME here. Eventually I will fix it myself when working on the hybrid

damencho: I don't see such method in WebDriver interface, seems it is a mobile specific. I don't even see it in the Appium interfaces.
But any request will do the work ...

PR #111, opened 2018-01-11
Add chat panel tests

bgrozev:
I had a discussion with Damian about adding these new methods to Participant, and we didn't reach a decision. I prefer this way rather than using a utility, but I don't have a strong preference (and there are other solutions as well). I also don't know how this will play with potentially using a page object approach (although I imagine these high level methods such as setDisplayName can be implemented on top of page objects). @paweldomas, please let us know what you think. I'd be happy to move the methods to a utility class or whatever you guys decide.


paweldomas:
Yeah, it's much better to have those methods in Participant than in utility methods IMO.

src/test/java/org/jitsi/meet/test/ChatPanelTest.java
```diff
@@ -0,0 +1,112 @@
...
+            isChatPanelOpen(),
+            "The chat panel should be closed after clicking the button");
+    }
+
+    /**
+     * Checks whether the chat panel is currently open/visible.
+     * @return {@code true} if the chat panel is open/visible, and {@code false}
+     * otherwise.
+     */
+    private boolean isChatPanelOpen()
```
paweldomas: I think we could move those 3 methods: isChatPanelOpen, clickChatPanelButton and pressChatPanelShortcut to a ChatPanel PageObject.

src/test/java/org/jitsi/meet/test/web/WebChatPanel.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public void clickToolbarButton()
+    {
+        MeetUIUtils.clickOnToolbarButton(
+            participant.getDriver(),
+            "toolbar_button_chat",
+            false);
+    }
+
+    @Override
```
paweldomas: Actually could you copy the JSDocs from the old methods ?

PR #118, opened 2018-02-07
feat(info): new tests for invite and locking

virtuacoplenny:
Relies on https://github.com/jitsi/jitsi-meet-torture/pull/119 being merged in first. This PR will need to be rebased after that PR is merged.


src/test/java/org/jitsi/meet/test/web/DialInNumbersPage.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Represents the dial in numbers page in a particular {@link WebParticipant}.
+ *
+ * @author Leonard Kim
+ */
+public class DialInNumbersPage
```
paweldomas: Do you think it would make sense to put page objects into another folder ? For example pageobjects/base, pageobjects/web, pageobjects/mobile. So far we have only ChatPanel so it should go there as well if we decide to do that.

virtuacoplenny: I do think it makes sense to have a separate folder for page objects that clearly implies or explicitly states the folder is for page objects, yes. To me that makes sense from an organizational standpoint, compared to what exists right now which is tests.web and test.mobile. Would the structure be test.pagobjects with subfolders web and mobile for now? That seems reasonable.

paweldomas: yes

virtuacoplenny: Okay, I'm gonna make a separate PR for moving some files around. These new tests will depend on that PR. Once that PR is in, I'll rebase and move around my new page objects.

virtuacoplenny:
Force pushed to put the new page objects into the new page objects package.

PR #121, opened 2018-02-12
feat(SS): Implement test

hristoterezov:


src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
```diff
@@ -178,11 +192,25 @@ else if (participantType == ParticipantType.edge)
...
+            if(options.getChromeExtensionId() != null)
+            {
+                try
+                {
+                    ops.addExtensions(
+                        downloadExtension(options.getChromeExtensionId()));
+                }
+                catch (IOException e)
+                {
+                    e.printStackTrace();
```
paweldomas: what ? shouldn't that fail the test ?

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
```diff
@@ -322,4 +350,24 @@ private URL getRemoteDriverAddress()
...
+    /**
+     * Downloads extension from chrome webstore.
+     * @param id - the id of the extension.
+     * @returns <tt>File</tt> associated with the downloaded extension.
+     */
+    private static File downloadExtension(String id)
+        throws IOException
+    {
+        URL website = new URL(
+            EXT_DOWNLOAD_URL_PREFIX + id + EXT_DOWNLOAD_URL_SUFIX);
```
paweldomas: why not user string format here ?

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
```diff
@@ -322,4 +350,24 @@ private URL getRemoteDriverAddress()
...
+        ReadableByteChannel rbc = Channels.newChannel(website.openStream());
+        File extension = File.createTempFile("jidesha", ".crx");
+        extension.deleteOnExit();
+
+        FileOutputStream fos = new FileOutputStream(extension);
```
paweldomas: use https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/test/java/org/jitsi/meet/test/util/MeetUtils.java
```diff
@@ -123,6 +123,22 @@ public static String getResourceJid(WebDriver participant)
...
+     * @param participant the <tt>WebDriver</tt> instance which runs conference
+     *                    participant.
+     * @param key the <tt>String</tt> key from config.js.
+     * @return the value for the given <tt>key</tt> from the config.js loaded
+     * for the given <tt>participant</tt>.
+     */
+    public static Object getConfigValue(WebDriver participant, String key)
+    {
+        return ((JavascriptExecutor) participant)
+                .executeScript("return config." + key);
```
paweldomas: I don't like this approach for getting config parameters, because it requires alive and connected to an existing conference WebParticipant. I don't see any of such assumptions being mentioned anywhere around here.

paweldomas:
Why don't we go with PageObjects given the test is being rewritten ?

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
hristoterezov: It will eventually fail the test probably anyway. How do you want to change it?

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
hristoterezov: How? My Java is a little rusty... What will be the benefit?

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
hristoterezov: Sorry, my java is a little bit rusty as I said. So could you please explain with more details? The part that puzzles me is I want this method to trow the exception.

src/test/java/org/jitsi/meet/test/util/MeetUtils.java
hristoterezov: OK. Could you be more specific which approach you would like for this situation? If you don't have better approach in mind, the current approach will be still the best....

hristoterezov:
> Why don't we go with PageObjects given the test is being rewritten ?

I have no idea what are you talking about! What is `PageObjects`?

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
paweldomas: to wrap in RuntimeException at least

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
paweldomas: yes, but the purpose of using that is to close the resource, not to catch the IOException

src/test/java/org/jitsi/meet/test/util/MeetUtils.java
paweldomas: Probably at least moving this method to WebParticipant, because having it there means that it belongs to web only.

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
paweldomas: https://dzone.com/articles/java-string-format-examples

It is weird that there are two constants for something which is actually one thing, where prefix is never used without the suffix.

It's possible that the % signs in the string would have to be escaped.

paweldomas:
PageObjects are used here for example: https://github.com/jitsi/jitsi-meet-torture/blob/master/src/test/java/org/jitsi/meet/test/ChatPanelTest.java

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
hristoterezov: OK. I'll read the docs more carefully!

src/test/java/org/jitsi/meet/test/util/MeetUtils.java
hristoterezov: OK

src/test/java/org/jitsi/meet/test/VideoLayoutTest.java
```diff
@@ -54,63 +54,55 @@ public void setupClass()
...
     }
 
     /**
      * The webdriver to test.
      * @param participant to test.
+     * @param isScreenSharing <tt>true</tt> if SS is started and <tt>false</tt>
+     *                        otherwise.
      */
-    void driverVideoLayoutTest(Participant participant)
+    void driverVideoLayoutTest(Participant participant, boolean isScreenSharing)
```
bgrozev: I think this method is now superfluous (and the doc needs an update).

damencho:
I've added pr testing for torture now. And this one failed:
unknown error: cannot process extension #1
from unknown error: cannot extract magic number
  (Driver info: chromedriver=2.35.528139 (47ead77cb35ad2a9a83248b292151462a66cd881),platform=Linux 4.4.0-1050-aws x86_64) (WARNING: The server did not provide any stacktrace information)

We should handle the case if desktop sharing is not configured.

src/test/java/org/jitsi/meet/test/VideoLayoutTest.java
hristoterezov: The docs are updated! I decided to leave it because in future we may want to add more stuff in it.

hristoterezov:
@paweldomas I don't see why we need to implement page object for tests that are running only on web. Especially for the desktop sharing test, we don't have any idea if this feature will ever be implemented for mobile and how it will look there. Even if we refactor it now we'll have to definitely refactor it later again when the functionality is implemented for mobile. I think that for desktop sharing test adding page objects will be waste of time (3 times refactoring/rewriting). At the end if there is a desktop sharing for mobile, it can be so different that we may even need totally different test.

I'm sorry if I misunderstood the idea behind the page objects! Please correct me if I'm wrong!

paweldomas:
The ChatPanel test is currently only for web too, so I don't find this argument relevant. I will try to be more specific and comment on each place where that would be suitable and we can continue the discussion from there.

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
```diff
@@ -37,128 +31,76 @@
...
     {
-        return true;
+        super.setupClass();
+
+        ensureOneParticipant();
+
+        ensureTwoParticipants(
+            null, null, null,
+            new WebParticipantOptions().setChromeExtensionId(
+                (String) MeetUtils.getConfigValue(
```
paweldomas: Can we move getConfigValue method to WebParticipant ? Then you can call getParticipant1().getConfigValue(). Actually it could event be a method "getDesktopSharingExtensionId()" in the WebParticipantDirectly.

paweldomas: The reason I ask for that is that we're trying to go away from accessing the driver directly and using it with static utility functions.

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
```diff
@@ -37,128 +31,76 @@
...
-            = remoteParticipantID.replaceAll("participant_", "");
+        String participant1Jid
+            = MeetUtils.getResourceJid(getParticipant2().getDriver());
 
-        final String expectedResult = "desktop";
         // holds the last retrieved value for the remote type
         final Object[] remoteVideoType = new Object[1];
         try
         {
             final String scriptToExecute = "return APP.UI.getRemoteVideoType('"
```
paweldomas: can we make this script execution a "getRemoteVideoType" method in WebParticipant ?

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
```diff
@@ -167,20 +109,69 @@ public void testDesktopSharingInPresence()
...
+    {
+        // check layout
+        new VideoLayoutTest().driverVideoLayoutTest(
+            getParticipant1(), isScreenSharing);
+
+        // hide thumbs
+        MeetUIUtils.clickOnToolbarButton(
+            getParticipant1().getDriver(), "toggleFilmstripButton");
+
+        TestUtils.waitMillis(5000);
```
paweldomas: What are we waiting here for exactly ? Waits like this are waste of time and never check for the condition that is expected after the wait period is over.

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
```diff
@@ -167,20 +109,69 @@ public void testDesktopSharingInPresence()
...
+
+        // check layout
+        new VideoLayoutTest().driverVideoLayoutTest(
+            getParticipant1(), isScreenSharing);
+
+        // show thumbs
+        MeetUIUtils.clickOnToolbarButton(
+            getParticipant1().getDriver(), "toggleFilmstripButton");
+
+        TestUtils.waitMillis(5000);
```
paweldomas: Again: what's the purpose of the wait here ?

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
```diff
@@ -167,20 +116,69 @@ public void testDesktopSharingInPresence()
...
-        {
-            e.printStackTrace();
-        }
-        ensureTwoParticipants();
+    /**
+     * Starts desktop sharing.
+     */
+    private void startDesktopSharing()
+    {
+        TestUtils.waitForElementByXPath(
```
paweldomas: So this 3 statements in the PageObject world would be:
Toolbat toolbar = getParticipant2().getToolbar();
Button dsButton = toolbar.getDesktopsharingButton();
dsButton.click();
dsButton.assertToggled(true /* toggled */, 2 /* timeout seconds */);
I don't know what would be the exact type of the Button. Maybe it's just WebElement. Or maybe we need a class the wraps it.

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
```diff
@@ -167,20 +116,69 @@ public void testDesktopSharingInPresence()
...
+                getParticipant2().getDriver(), DS_BUTTON_XPATH,
+                "toggled", 2);
+    }
+
+    /**
+     * Stops desktop sharing.
+     */
+    private void stopDesktopSharing()
+    {
+        TestUtils.waitForElementByXPath(
```
paweldomas: Same as above.

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
```diff
@@ -167,20 +116,69 @@ public void testDesktopSharingInPresence()
...
+    /**
+     * Checks the video layout on the other side, after we imitate
+     * desktop sharing.
+     * @param isScreenSharing <tt>true</tt> if SS is started and <tt>false</tt>
+     *                        otherwise.
+     */
+    private void checkExpandingDesktopSharingLargeVideo(boolean isScreenSharing)
+    {
+        // check layout
+        new VideoLayoutTest().driverVideoLayoutTest(
```
paweldomas: I hate that we need to create test instances somewhere else, but it would be probably too much to refactor this in this PR...

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
damencho: I think the approach with getButton and then click is not good at all in the context of react. I would even say to skip searching for web element and getting them and then executing click. We already had such problems with the toolbars. The better approach is at the end to just execute document.getElementById('...').click();
https://github.com/jitsi/jitsi-meet-torture/blob/master/src/test/java/org/jitsi/meet/test/util/MeetUIUtils.java#L243
The problem is a StaleElementException coming from selenium. You take the element and try to click it, but by that time react has re-rendered the toolbar and that element is stale.

bgrozev: Note that this is a problem with saving the WebElement, and not a problem with the "getButton" abstraction itself.

damencho: Yep, that's right, thanks for the clarification :)

hristoterezov:
@paweldomas Thanks very much for the additional explanations. Maybe I was not very clear. My problem here is that I don't think we should use the PageObjects for this test. But let's discuss this offline because it seems like a longer discussion.

paweldomas:
@hristoterezov Why would we have two different ways of dealing with things ? Everything is "longer discussion" always. I keep hearing that all the time recently. We can discuss that on the retro or call a meeting.

paweldomas:
@damencho we can do document.getElemntById on web inside of the Button or PageObject abstraction. But on mobile it will be a different way for getting WebElement.

paweldomas:
Btw. long time ago I was forbidden to rewrite whole torture into the PageObject concept and there was an agreement  to do that step by step, so here we are. I am really getting frustrated about this attitude of getting shortcuts all the time and then someone else when adding new thing will have to deal with all that. You have just a couple of places to address. I'm not asking you to write full Toolbar or whatever -just the part you need (1 button ?).

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
hristoterezov: See the previous comments. I've agreed with this and getConfigValue is already a moved to WebParticipant.

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
hristoterezov: Is there other place where we are using this? I don't think so. That's why I think this belongs to the DesktopSharingTest!

paweldomas: No, it does not ! It belongs to the WebParticipant !

paweldomas: Test class describes what is being tested and how. Who's video type is ? Participant's. In this particular case we're checking what video type given web participant has received for given remote participant through presence. Other tests can be also interested in checking that.

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
damencho: We will leave this for now like this. Hiding the filmstrip because of animations can take some time. We cannot check for a css change, cause currently, we change the css and after the change, it starts moving the filmstrip away from the screen for two seconds (which with the framebuffer can take more). We already spend a lot of time looking at it to figure it out. Maybe checking for isDisplayed of the local video is one solution will give it a quick try and if it doesn't work will leave the waits.

paweldomas: So the purpose could be explained in a comment at least.

damencho: Seems to wait for local video not displayed works. Will see how it goes. Will merge it after the tests pass.

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
```diff
@@ -134,6 +129,9 @@ private void checkExpandingDesktopSharingLargeVideo(boolean isScreenSharing)
         MeetUIUtils.clickOnToolbarButton(
             getParticipant1().getDriver(), "toggleFilmstripButton");
 
-        TestUtils.waitMillis(5000);
+        TestUtils.waitForDisplayedElementByXPath(
+            getParticipant1().getDriver(),
+            "//span[@id='localVideoWrapper']",
```
paweldomas: this duplicates the constant and could be a TestUtil method, but I give up since you want to merge it now

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
```diff
@@ -37,150 +29,115 @@
...
-        ensureOneParticipant();
+    /**
+     * Checks the status of desktop sharing received from the presence and
+     * compares it to the passed expected result.
+     * @param expectedResult camera/desktop
+     */
+    private void testDesktopSharingInPresence(final String  expectedResult)
+    {
+        String participant1Jid
+            = MeetUtils.getResourceJid(getParticipant2().getDriver());
```
bgrozev: The variable name suggests this is the id of participant 1, but the code gets the id of participant 2. What is the intention?

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
```diff
@@ -37,150 +29,115 @@
...
-        catch (TimeoutException e)
-        {
-            assertEquals(
-                expectedResult, remoteVideoType[0],
-                "Wrong video type, maybe desktop sharing didn't work");
-        }
+    /**
+     * Stops desktop sharing.
+     */
+    private void stopDesktopSharing()
```
bgrozev: startDesktopSharing and stopDesktopSharing are exactly the same? Is this intentional?

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
hristoterezov: No intention. It's a mistake! I should change the variable name!

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
hristoterezov: Kind of. After the refactoring I end up implementing a method that handles both cases. Maybe I should remove the start/stopDesktopSharing and use the new method directly?

src/test/java/org/jitsi/meet/test/VideoLayoutTest.java
bgrozev: > The docs are updated!

I don't think this actually made it to the PR: https://github.com/jitsi/jitsi-meet-torture/blob/925dec78ff39b10cfbe96e2c56b374ee693e63bd/src/test/java/org/jitsi/meet/test/VideoLayoutTest.java#L61 (this commit is the current head of master)

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
bgrozev: I'm merging anyway and replaced the two methods with "toggleDesktopSharing"

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
bgrozev: I'm merging anyway and will change it

PR #122, opened 2018-02-14
ref(lock-room): replace LockRoomTest with NewLockRoomTest

virtuacoplenny:
Now that Invite Modal is being removed, remove LockRoomTest
which exercised the modal's lock feature. Rename NewLockRoomTest
which also exercises the lock feature but in the Info Dialog.


virtuacoplenny:
force pushed to remove conflicts

PR #123, opened 2018-02-14
Records logs and screenshots when configuration fails.

damencho:
Moves the failed configuration tests to the set of failed tests to be easily recognised in html report.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
+++ b/src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
@@ -134,6 +134,9 @@ public abstract class AbstractBaseTest
 
         participants = new ParticipantHelper(new ParticipantFactory(config));
 
+        // if this one fails, the failure will be registered in the
+        // FailureListener to gather information and it will call
+        // the cleanupClass() to clean up any leftovers
         setupClass();
     }
 
--- a/src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
+++ b/src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
@@ -134,17 +134,7 @@ public abstract class AbstractBaseTest
 
         participants = new ParticipantHelper(new ParticipantFactory(config));
 
-        // make sure if setup fails we will cleanup
-        // when configure method @BeforeClass fails @AfterClass is not executed
-        try
-        {
-            setupClass();
-        }
-        catch (Throwable t)
-        {
-            cleanupClass();
-            throw t;
-        }
+        setupClass();
     }
 
     /**
--- a/src/test/java/org/jitsi/meet/test/base/FailureListener.java
+++ b/src/test/java/org/jitsi/meet/test/base/FailureListener.java
@@ -186,8 +186,12 @@ public class FailureListener
     @Override
     public void onConfigurationFailure(ITestResult itr)
     {
+        // we want to catch failures to configure a test, so we can record
+        // logs, screenshots and any information as we do as normal test fails
+        // and we skip any other configuration failures as @AfterClass
         Object testInstance = itr.getInstance();
-        if (testInstance instanceof AbstractBaseTest)
+        if (testInstance instanceof AbstractBaseTest
+            && itr.getMethod().isBeforeClassConfiguration())
         {
             // record whatever we can
             onTestFailure(itr);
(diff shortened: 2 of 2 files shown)


src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
```diff
@@ -134,17 +134,7 @@ final public void setupClassPrivate(ITestContext context)
...
-        try
-        {
-            setupClass();
-        }
-        catch (Throwable t)
-        {
-            cleanupClass();
-            throw t;
-        }
+        setupClass();
```
paweldomas: so when setupClass() fails will the cleanupClass() be called or not ?

paweldomas: sorry, I didn;t read the other code

src/test/java/org/jitsi/meet/test/base/FailureListener.java
```diff
@@ -114,7 +115,15 @@ public void onStart(ITestContext iTestContext)
 
     @Override
     public void onFinish(ITestContext iTestContext)
-    {}
+    {
+        // move all failed configurations to the failed tests set
```
paweldomas: so is it that failed configurations do not appear at all in the report ?

src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
paweldomas: It will be useful if we had a comment either here or in cleanupClass description that it will be called by FailureListener if that fails.

src/test/java/org/jitsi/meet/test/base/FailureListener.java
```diff
@@ -170,6 +179,30 @@ public void onTestFailure(ITestResult testResult)
         }
     }
 
+    @Override
+    public void onConfigurationSuccess(ITestResult itr)
+    {}
+
+    @Override
+    public void onConfigurationFailure(ITestResult itr)
```
paweldomas: This callback is called for every configuration method that fails including @AfterClass, @AfterSuite etc. Because our intention here is to handle only "setupClass" I wonder if it's possible to check for method name here.

It may turn out that it's better to keep try catch and cleanupClass where it was, because this logic is kind of specific to AbstractBaseTest. It knows that it has participant to be cleaned up. In my opinion it would be more readable to leave it there.

Even if you leave it where it was the new functionality about adding to failed results should work, because it throws the throwable (results in a failure).

src/test/java/org/jitsi/meet/test/base/FailureListener.java
damencho: Nope, its that we will see that test also as failed.

src/test/java/org/jitsi/meet/test/base/FailureListener.java
damencho: The problem is gathering data. Currently there is no screenshot or anything if that fail and we catch it in the AbstractBaseTest.

damencho: I will check whether we can limit it for just BeforeClass

damencho: It seems it can be easily done https://static.javadoc.io/org.testng/testng/6.8.21/org/testng/ITestNGMethod.html#isBeforeClassConfiguration()
I will test it.

src/test/java/org/jitsi/meet/test/base/FailureListener.java
```diff
@@ -170,6 +179,30 @@ public void onTestFailure(ITestResult testResult)
...
+        Object testInstance = itr.getInstance();
+        if (testInstance instanceof AbstractBaseTest)
+        {
+            // record whatever we can
+            onTestFailure(itr);
+
+            // make sure if setup fails we will cleanup
+            // when configure method @BeforeClass fails @AfterClass is not
+            // executed
+            ((AbstractBaseTest)testInstance).cleanupClass();
```
paweldomas: I understand we need to take the screenshots, logs, etc, but doing the cleanup should be part of AbstractBaseTest.

paweldomas: or is it that if we cleanup, there will be nothing to take the screenshots from ?

damencho: That's correct, we need to do it before cleaning

damencho:
Is this ok?

paweldomas:
yes, thanks !

PR #124, opened 2018-02-15
fix(VideoOperator): fix class cast exception

paweldomas:
The method sometimes fails with: "java.lang.ClassCastException:
java.lang.Long cannot be cast to java.lang.Double".

I guess that if the value happens to be integer it can be Long instead
of Double. Make it Number given that it's only used in a debug log.


paweldomas:
Actually it doesn't matter if this PR is merged, before or after fix for screenshots dir.

PR #125, opened 2018-02-15
Mkdirs

paweldomas:
Log error when mkdirs() fails.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/base/TestSettings.java
+++ b/src/test/java/org/jitsi/meet/test/base/TestSettings.java
@@ -61,29 +61,11 @@ public class TestSettings
             // setting system properties
             System.setProperties(p);
 
-            initFolders();
-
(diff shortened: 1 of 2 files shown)


src/test/java/org/jitsi/meet/test/base/TestSettings.java
```diff
@@ -61,27 +61,11 @@ public static Properties initSettings()
             // setting system properties
             System.setProperties(p);
 
-            initFolders();
-
```
damencho: Don't we need appium screenshots folder creation?

paweldomas: it is done as part of https://github.com/jitsi/jitsi-meet-torture/blob/master/src/test/java/org/jitsi/meet/test/base/FailureListener.java#L63

PR #129, opened 2018-02-16
Removes ParticipantFactoryConfig in favor of ParticipantOptions.

paweldomas:
I'm opening this PR to be able to put comments on the code


src/test/java/org/jitsi/meet/test/ConferenceMigrationTest.java
```diff
@@ -94,10 +94,12 @@ public void testConferenceMigration()
             "Migrated bridge: " + migratedBridge +
             ", REST endpoint: " + jvbRESTEndpoint);
 
-        ensureOneParticipant("config.enforcedBridge=\"" + migratedBridge +"\"");
-        Participant owner = getParticipant1();
+        ensureTwoParticipants(
```
paweldomas: Because of "ensureTwo/ThreeParticipants" methods we have so many troubles with overloads. Once will have only URL, the other time it will also have participant options. What if we start adding "ensureFour/Five/SixParticipants" ? I think we would have less problems and the code would be simpler if we'd change those to:
`
Participant p1 = createParticipant(0 /* index/number */, participantOptions);
Participant p2 = createParticipant(1, participantOptions);
p1.join(url);
p2.join(url);
`
To make porting easier we could leave "ensureOne/Two/Three", but only if there are no arguments. WDYT ?

src/test/java/org/jitsi/meet/test/base/ParticipantOptions.java
```diff
@@ -0,0 +1,148 @@
...
+{
+    public static final String FAKE_AUDIO_FNAME_PROP
+        = "jitsi-meet.fakeStreamAudioFile";
+
+    public static final String FAKE_VIDEO_FNAME_PROP
+        = "jitsi-meet.fakeStreamVideoFile";
+
+    private String configPrefix;
+
+    public static enum Options
```
paweldomas: Most of those options are web specific and we can not have enum subclasses. We need to figure this out in some other way... I was thinking that options should be backed up by a Properties object and we'd have ParticipantOptions subclasses that declare keys ?

paweldomas: Also I was thinking we should have WebParticipantBuilder like there are mobile ones and it would provide default values. Plus the loading should happen in ParticipantFactory which has the config object and it could read from there instead of System.getProperty (which is contained in the config).

paweldomas: Or a WebParticipantFactory would provide default WebParticipantOptions. It's kind of hard to guess the details without trying that out.

src/test/java/org/jitsi/meet/test/ConferenceMigrationTest.java
paweldomas: Also we should think of a proper way for dealing with "joinRef". We could have new "web participant option" which tells WebParticipant that it should use another way to join the conference.

src/test/java/org/jitsi/meet/test/base/Participant.java
```diff
@@ -115,30 +99,26 @@ public Participant(
...
-        JitsiMeetUrl conferenceUrl = (JitsiMeetUrl) this.meetURL.clone();
-
-        conferenceUrl.setRoomName(roomName);
-        conferenceUrl.setRoomParameters(roomParameter);
-        conferenceUrl.appendConfig(config);
+        meetURL = Objects.requireNonNull(meetURL, "meetURL");
+        if (meetURL.getRoomName() == null)
+        {
+            meetURL.setRoomName(roomName);
+        }
```
damencho: This check for not overriding the room name is needed for the iframetest. Otherwise it overrides the roomname and breakes the iframe link and we see access denied.

paweldomas: can we get rid of the roomName argument in joinConference ?

damencho: Yep, I was thinking the same, but didn't have time to try it. I suppose we can get rid of it :)

src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
```diff
@@ -177,6 +184,23 @@ private Properties mergeTestNGSuiteProperties(
...
+     * {@link JitsiMeetUrl#serverUrl} field initialized with the value from
+     * {@link #JITSI_MEET_URL_PROP} system property.
+     *
+     * @return a new instance of {@link JitsiMeetUrl}.
+     */
+    public JitsiMeetUrl getJitsiMeetUrl()
+    {
+        JitsiMeetUrl url = new JitsiMeetUrl();
+
+        url.setServerUrl(System.getProperty(JITSI_MEET_URL_PROP));
```
paweldomas: can we use config instead of System.getProperty ? Eventually this could come from the factory which stores the config already.

src/test/java/org/jitsi/meet/test/web/WebTestBase.java
```diff
@@ -258,6 +301,23 @@ private Participant joinParticipant(
...
+    }
+
+    /**
+     * Return new {@link JitsiMeetUrl} instance which has only
+     * {@link JitsiMeetUrl#serverUrl} field initialized with the value from
+     * {@link #JITSI_MEET_URL_PROP} system property.
+     *
+     * @return a new instance of {@link JitsiMeetUrl}.
+     */
+    @Override
```
paweldomas: If we'd want to append default config for mobile we'll have to override this method and duplicate the logic. If default config would be a param to base class then it could do it. Anyway it probably fits better to Participant, because a test may want to have both mobile and web participants. Plus this config part is specific to web/mobile participant and it feels like it belongs to Participant.

src/test/java/org/jitsi/meet/test/web/WebTestBase.java
```diff
@@ -238,7 +270,12 @@ private Participant joinParticipant(
 
             String configPrefix = "web.participant" + (index + 1);
 
-            p = participants.createParticipant(configPrefix);
+            if (options == null)
```
paweldomas: The logic for providing default ParticipantOptions should be in shared code. One way fo doing it would be to move it to the ParticipantHelper.createParticipant and allow null argument.

src/test/java/org/jitsi/meet/test/base/ParticipantOptions.java
```diff
@@ -0,0 +1,60 @@
...
+     *
+     */
+    public void load(String configPrefix)
+    {
+        this.configPrefix = configPrefix;
+
+        // It will be Chrome by default...
+        ParticipantType participantType
+            =  ParticipantType.valueOfString(
+                    System.getProperty(configPrefix + ".type"));
```
paweldomas: use config instead of System.getProperty ?

src/test/java/org/jitsi/meet/test/web/WebParticipantOptions.java
```diff
@@ -0,0 +1,111 @@
...
+        String fakeStreamVideoFile)
+    {
+        this.put(FAKE_VIDEO_PROP, fakeStreamVideoFile);
+
+        return this;
+    }
+
+    public boolean getBoolean(Object key)
+    {
+        Object v = this.get(key);
```
paweldomas: this can be simplified to:
`return v != null && (boolean) v;`

plus as a utility getter probably fits better in the base class.

src/test/java/org/jitsi/meet/test/web/WebParticipantOptions.java
```diff
@@ -0,0 +1,111 @@
...
+    {
+        super(config);
+
+        initDefaults();
+    }
+
+    private void initDefaults()
+    {
+        String fakeStreamAudioFile = config.getProperty(FAKE_AUDIO_FNAME_PROP);
+        if (fakeStreamAudioFile == null)
```
paweldomas: For 'fakeStreamVideoFile' it also checks for blank, but here only for null. How about using org.apache.commons.lang3.StringUtils.isNotBlank for both ?

src/test/java/org/jitsi/meet/test/web/WebParticipantOptions.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    @Override
+    public void load(String configPrefix)
+    {
+        super.load(configPrefix);
+
+        String isRemote = config.getProperty(configPrefix + ".isRemote");
+        if (isRemote != null)
```
paweldomas: These checks for null are duplicated logic. How about adding utility method, something like loadConfigProperty(configPrefx, propertyKey) ?

src/test/java/org/jitsi/meet/test/web/WebParticipantOptions.java
```diff
@@ -0,0 +1,111 @@
...
+            this.put(REMOTE_PROP, Boolean.valueOf(isRemote));
+
+        String browserBinary = config.getProperty(configPrefix + ".binary");
+        if (browserBinary != null)
+            this.put(BINARY_PROP, browserBinary);
+
+        String version = config.getProperty(configPrefix + ".version");
```
paweldomas: can we use VERSION_PROP constant instead of ".version" ? It could be done automagically in the "loadConfigProperty" mentioned above. Btw. is there any reason why VERSION_PROP is different case that the constant ? I guess they could be the same.

src/test/java/org/jitsi/meet/test/web/WebParticipantOptions.java
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Sets the name of wav audio file which will be streamed through fake audio
+     * device by participants. The file location is relative to working folder.
+     * For remote drivers a parent folder can be set and the file will be
+     * searched in there.
+     *
+     * @param fakeStreamAudioFile full name of wav file for the fake audio
+     *                            device.
+     */
+    public WebParticipantOptions setFakeStreamAudioFile(
```
paweldomas: do we want to add get/set method for every property given that ParticipantOptions extends Properties which basically provides direct access anyway ? I think it would be better to not extend Properties, but have it as a backed in "properties" field. Then custom get/put methods which could return "this" for chaining calls.
Or we could agree on having getters/settters for everything if it's more convenient, but then direct access should not be allowed and Properties should still be used only as a backend.

src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
```diff
@@ -107,6 +112,8 @@ protected AbstractBaseTest(AbstractBaseTest baseTest)
         participants = new ParticipantHelper(baseTest.participants);
     }
 
+    public abstract ParticipantFactory getFactory(Properties config);
```
paweldomas: Previous code was selecting factory based on the participant type. This way a test was able to create both mobile and web participants. That's not possible anymore.

paweldomas:
That's all comments for now. Please let me know what do you think.

src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
damencho: Ok, I see what you mean. If we separate stuff and we have web and mobile extending base, I think it's good to not import web or mobile in base. I will look for solution of this.

src/test/java/org/jitsi/meet/test/web/WebParticipantOptions.java
damencho: Getter and setter is convenient for stuff we use a lot. Having an option to add and get custom stuff is more convenient to be able to easily add stuff without the need to modify options class, but can lead to props scattered around ... Not sure what is the best way to do it ...

src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
paweldomas: I don't see why this would be a problem here. It's not Jitsi OSGi bundle nor we're looking to have any plugin system here, but I'm curious what will you come up with.

src/test/java/org/jitsi/meet/test/base/ParticipantOptions.java
```diff
@@ -55,12 +34,18 @@ static ParticipantType getParticipantType(
...
+        String configPrefix,
+        ParticipantOptions overrides)
     {
-        this.configPrefix = configPrefix;
+        this.setProperty(TYPE_PROP, getParticipantType(config, configPrefix));
 
         String name = configPrefix.substring(configPrefix.indexOf('.') + 1);
         this.setProperty(NAME_PROP, name);
+
+        if (overrides != null)
```
paweldomas: Can we add "override" or "merge" method and move the override there ? Those overrides normally should be of no interest to the load method (to not make things more complex and encourage to write some weird stuff here). I see few other things that can be simplified, but I don't want to torture you with that anymore. I'll do that in a separate PR eventually. If the tests will pass it should be good enough for now. But please remember about missing JavaDocs...

damencho: Check it now, hope all javadocs are there. The tests pass (just the LockRoom but it is fixed in master).

src/test/java/org/jitsi/meet/test/web/WebParticipantOptions.java
```diff
@@ -0,0 +1,143 @@
...
+     * A prefix for global options (not per participant).
+     */
+    private static final String GLOBAL_PROP_PREFIX = "jitsi-meet";
+
+    /**
+     * Constructs new WebParticipantOptions, filling it with default values.
+     */
+    public WebParticipantOptions()
+    {
+        this.setProperty(FAKE_AUDIO_PROP, "resources/fakeAudioStream.wav");
```
paweldomas: should there be also a default for FAKE_VIDEO_PROP (I don't remember if we've had one) ?

paweldomas: There will be a problem with setting default in the constructor when merging explicit ParticipantOptions on top of config loaded options (it will always override anything there). I think we should have a method somewhere (maybe in factory) which gets the default ParticipantOptions set and then all 3 options sets are being merge one on top of another. I'm looking into that now how that would looks like and if that make sense, as I'm not sure.

damencho: Nope, we only have one for lipsync and psnr.

damencho: But wait, these are defaults so if there are no settings in override this will stay. If there is something in those that will go on top of this, this is what we want. I think I don't get the problem.

paweldomas: Yes, actually it will work this way. Forget about that

PR #132, opened 2018-02-16
fix(info): for now avoid chained look ups to avoid stale elements

virtuacoplenny:
A better solution has to be thought of and implemented. For now
go with the approach of not chaining in case the element has
become stale from a react re-render.


virtuacoplenny:
The tests passed 10 times in a row for me against a selenium grid. Locally I couldn't reproduce the issue before changing the lookups.

PR #134, opened 2018-02-22
Skips testEmailPersistence on FF participant.

damencho:


src/test/java/org/jitsi/meet/test/AvatarTest.java
```diff
@@ -201,6 +201,14 @@ public void avatarWhenVideoMuted()
     @Test(dependsOnMethods = { "avatarWhenVideoMuted" })
     public void testEmailPersistence()
     {
+        // This test is very often failing on FF (due to a crash on hangup
+        // present in FF57)
+        // will disable it for now
+        if (getParticipant1().getType() == ParticipantType.firefox)
+        {
+            return;
+        }
```
virtuacoplenny: Is this the standard way to disable a test for a browser or at least the expected way in jitsi-meet-torture? There isn't a configuration you can set somewhere else so the tests don't have to know about this special case?

damencho: I want to disable just that method, I'm not sure there is a configuration about that. Of course we can scrap this one, and just disable all AvatarTest when starting FF testing, but I didn't want to skip the other part just because of FF webrtc bug. There can probably be another workaround like adding a 15-second wait before starting this test method, I suspect the problem comes from the fact that video was unmuted recently, but I don't have time to experiment with that and the fact this is really random one I need 20 runs to verify it.
And this is the safest way for the moment to not run it, as people will start complaining again.

damencho: But if we do not merge this I will definitely just disable the AvatarTests for FF and we need to enable them once FF fixes all their issues. Currently, 58 and 59 also do not work because of getUserMedia never returning ... and not sure about the webrtc bug cause tests don't even start :) without getUserMedia

paweldomas: eh... so we currently run the tests on FF 57 and then 58 and 59 will not work at all ?

damencho: Yep, that is correct.
https://bugzilla.mozilla.org/show_bug.cgi?id=1431549
https://bugzilla.mozilla.org/show_bug.cgi?id=1434477

Thanks for merging it.

damencho: The getUsermedia thing seems to be fixed in 60 :)

PR #135, opened 2018-02-26
feat: Adds a test for bandwidth estimations

gpolitis:
This is a WIP. Please don't merge yet. It depends on https://github.com/jitsi/lib-jitsi-meet/pull/713 and https://github.com/jitsi/jitsi-meet/pull/2535.


src/test/java/org/jitsi/meet/test/base/FailureListener.java
```diff
@@ -50,7 +50,7 @@
      * system property.
      * @return
      */
-    private static String getReportFolder()
+    public static String getReportFolder()
     {
```
damencho: You don't need that anymore, can you revert this little change.

damencho:
Can you also rename tc-port-sch.sh to tc-port-scheduler.sh, or whatever is the appropriate name, as we talked offline, thanks.

PR #136, opened 2018-02-27
fix: Another attempt to fix ChatPanelTest.

bgrozev:


bgrozev:
> moveToElement doesn't work in Firefox.

Do you know of an alternative which works?

damencho:
No, sorry. I was experimenting with something like:
executeScript("document.getElementById('largeVideo').focus();");

PR #139, opened 2018-03-06
Forbid TURN for the receive participant in the BW estimation tests.

gpolitis:


bgrozev:
The first 3 commits are already in master as part of #135. Rebase the last commit on top of the current master?

PR #142, opened 2018-03-07
feat: Debuggable bandwidth estimation tests.

gpolitis:


scripts/push-bwe-benchmark.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh -e
+
+readonly URL=$1
+for i in target/chrome-chrome-reports/logs/*.out;
+do
```
damencho: I started wondering about this chrome-chrome-reports, as this is a configurable name and can change.
But why do we want this as part of the source-code, this is a deployment specific ...

PR #143, opened 2018-03-07
ref(Participant): mobile does not execute scripts

paweldomas:
Mobile is not capable of executing any JavaScript code and probably never will be, so move away any code that depends on that to the WebParticipant.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/web/WebParticipant.java
+++ b/src/test/java/org/jitsi/meet/test/web/WebParticipant.java
@@ -166,10 +166,7 @@ public class WebParticipant extends Participant<WebDriver>
     @Override
     public Object executeScript(String var1, Object... var2)
     {
-        JavascriptExecutor executor = getJSExecutor();
-
-        return executor != null
-            ? executor.executeScript(var1, var2) : null;
+        return getJSExecutor().executeScript(var1, var2);
     }
 
     /**
@@ -180,16 +177,20 @@ public class WebParticipant extends Participant<WebDriver>
     @Override
     public Object executeAsyncScript(String var1, Object... var2)
     {
-        JavascriptExecutor executor = getJSExecutor();
-
-        return executor != null
-            ? executor.executeAsyncScript(var1, var2) : null;
+        return getJSExecutor().executeAsyncScript(var1, var2);
     }
 
     private JavascriptExecutor getJSExecutor()
     {
-        return driver instanceof JavascriptExecutor
-            ? (JavascriptExecutor) driver : null;
+        if (driver instanceof JavascriptExecutor)
+        {
+            return (JavascriptExecutor) driver;
+        }
+        else
+        {
+            throw new RuntimeException(
+                "The driver is not capable of executing JavaScript");
+        }
     }
 
     /**
--- a/src/test/java/org/jitsi/meet/test/web/WebParticipant.java
+++ b/src/test/java/org/jitsi/meet/test/web/WebParticipant.java
@@ -237,12 +237,6 @@ public class WebParticipant extends Participant<WebDriver>
     @Override
     public String getProtocol()
     {
-        WebDriver driver = getDriver();
-
-        if (driver == null)
-        {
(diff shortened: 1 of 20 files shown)


bgrozev:
Some of the tests now depend on WebParticipant, but they don't seem intrinsically web-only (e.g. OneOnOneTest, DisplayNameTest, FollowMeTest). Does that mean that we never intend to have them support mobile, or that we'll need to update them in order to support mobile?

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
```diff
@@ -157,6 +159,126 @@ protected void doHangUp()
...
+     * Executes a script in this {@link WebParticipant}'s {@link WebDriver}.
+     * See {@link JavascriptExecutor#executeScript(String, Object...)}.
+     */
+    @Override
+    public Object executeScript(String var1, Object... var2)
+    {
+        JavascriptExecutor executor = getJSExecutor();
+
+        return executor != null
+            ? executor.executeScript(var1, var2) : null;
```
bgrozev: I know this is just code being moved, but it looks like it could swallow errors. The "driver" field is final and non-null, so "executor == null" only happens if the driver is not an instance of JavascriptExecutor. If this happens for WebParticipant, we should throw an exception and not silently return null.

paweldomas: true - will have this fixed

paweldomas:
That's right it makes them dependent, but that because of the way they were written. They must be updated in order to be platform agnostic by getting rid of "executeScript" calls. It's probably, because Participant's interface was misleading.

Please also note that those tests extend WebTestBase which makes them web tests only.

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
```diff
@@ -157,6 +159,126 @@ protected void doHangUp()
...
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getProtocol()
+    {
+        WebDriver driver = getDriver();
```
bgrozev: Can we remove this unnecessary check for "driver"?

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
```diff
@@ -157,6 +159,126 @@ protected void doHangUp()
...
+    @Override
+    public boolean isP2pConnected()
+    {
+        // FIXME hmm this checks for ICE connected state which does not
+        // necessarily mean that the conference is currently in the P2P mode.
+        // It may turn out we're not really checking if P2P is connected
+        // and the method name may be confusing.
+        Object result = executeScript(MeetUtils.ICE_CONNECTED_CHECK_SCRIPT);
+
+        return Boolean.valueOf(result.toString());
```
bgrozev: haha, each of these very similar methods is implemented differently (one does a toString, one casts to String, one does a try/catch). :)
This one seems prone to throw a NPE if `result` is null. Just an observation, not a critique of this PR, and I don't expect a fix.

paweldomas: I will fix that. Kind of wasn't focused on those details more on moving the code around. Thanks !

paweldomas: I'm thinking what's the best way. How about this ?
return result != null && Boolean.valueOf(result.toString());
(possible wrapped into a method)

bgrozev:
Hmm, I see. Not related to this PR: so if we want tests to be running on both web and mobile we should avoid using WebTestBase and MobileTestBase? Should we also aim to move code from WebTestBase to AbstractBaseTest?

paweldomas:
That's right. Now I'm working on a HybridTest class which is capable of dealing with either type of Participants. Maybe it will be just the AbstractBaseTest if I manage to clean up the details.

PR #147, opened 2018-03-09
Enables RemoteDriver for safari.

damencho:


src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
```diff
@@ -158,6 +158,15 @@ private WebDriver startWebDriver(
         }
         else if (participantType == ParticipantType.safari)
         {
+            // You must enable the 'Allow Remote Automation' option in
+            // Safari's Develop menu to control Safari via WebDriver.
+            // In Safari->Preferences->Websites, select Camera,
+            // and select Allow for "When visiting other websites"
+            if (isRemote)
```
paweldomas: The isRemote condition seems to be special case for each of the if branches, but I guess it's not easy to be optimised at this point

PR #148, opened 2018-03-09
Fixes assertEquals params order, actual is first and second is expected.

damencho:


damencho:
Nope, I think we introduced them when we moved to testng.

PR #149, opened 2018-03-09
Adds rtp stats json, to the report on failures.

damencho:


src/test/java/org/jitsi/meet/test/util/MeetUtils.java
```diff
@@ -76,7 +76,8 @@
      * of the specified peer connection.
      *
      * @param driver
-     * @param useJVB TODO DOCUMENT
+     * @param useJVB true if we want the stats when the used connection was
+     * through jvb, or false if p2p is used.
      */
     public static String getRtcStats(WebDriver driver, boolean useJVB)
```
paweldomas: why is this method called getRtcStats and the other one getRtpStats ? Can we rename it (this one) ? Btw. this method should probably be in WebParticipant, but it's out of the scope of this PR...

damencho:
I renamed it and force pushed.

PR #150, opened 2018-03-09
Mobile participant options

paweldomas:
Refactors things to have MobileParticipantOptions instead of "builders" and to not import mobile and web "impl" packages from base.


bbaldino:
what advantage does this have over builders?  i find the builder syntax to be pretty nice in general, and the place where i looked at the old builder code vs the new one, the new version was way longer.

paweldomas:
I'm not saying it does have any advantage it's just what we've had for web and to make web use builders would be a lot of effort. Also there's no reason to have two different ways of dealing with things, so I wanted to adjust the mobile part. It's kind of hard to imagine for me these builders being used with the current web tests.

bbaldino:
personally i'd rather keep the builders (they seem better/cleaner to me, but maybe i'm wrong as i haven't been in this code much lately) and have a task to move the web in that direction if we think they're better.

> to make web use builders would be a lot of effort

you guys have done a lot of (awesome) refactoring for the tests, i'm not saying we'd need to make the builder change now, but this seems like making things worse for the sake of consistency (and doesn't seem like it was that small of a change in itself).

> It's kind of hard to imagine for me these builders being used with the current web tests.

why is that?  do you mean a builder pattern in general, or these ones specifically?

paweldomas:
>It's kind of hard to imagine for me these builders being used with the current web tests.

>>why is that? do you mean a builder pattern in general, or these ones specifically?

The goal behind this refactoring was to cleanup some inconsistencies in preparation for the hybrid test (a generic test which can be run agains web/mobile participants at the same time). By "hard to imagine" I meant more that it's impossible I will get time or be able to explain the need to spend time on refactoring the web tests to get the builders there. So I kind of needed one way to be able to work on those generic tests. Maybe it was just in my mind that I need one way, but it was making things easier to deal with.

paweldomas:
Actually I don't know how the tests would look like with the builders on web. Do you mean they should replace ParticipantOptions ? this syntax is not far from the builders

paweldomas:
Maybe it would be easier to discuss that on a call if you think it's worth the effort.

bbaldino:
> Actually I don't know how the tests would look like with the builders on web. Do you mean they should replace ParticipantOptions ? this syntax is not far from the builders.
> The problem is that there's a bunch of "ensureXParticipants" method which are kind of declarative, they do not expect to create/build a Participant instance and operate on it. They just throw some options there and expect the participants are there (then eventually they are obtained by getParticipant1(), getParticipant2()). There's a lot of code to be modified.

I may not have much value here because I haven't been in this code for a while, but I'm not sure I follow.  The builder is just used when creating a `Participant`, so I'd think that should just be in one place.  `getParticipantX()` will just be finding an already-created `Participant`, no?  How does how it is created impact that?  I tried taking a look at `getParticipant1()` and traced it back to its use of `ParticipantFactory`, which is what uses `ParticipantOptions`, so it actually seems like changing the creation of a `Participant` would only affect that one place if things are using the factory.

I don't quite get `ParticipantOptions` though, so I'm probably missing something there.

paweldomas:
You're right that in the factory it's kind of isolated, but there must be a generic way of passing the participant options in order to have a generic test. So we either need a generic builder or ParticipantOptions interface to be passed by the test. The ParticipantOptions describe what kind of participant will be created: is it chrome, firefox, android, ios ? is it local or does it run on selenium grid etc.

bbaldino:
> You're right that in the factory it's kind of isolated, but there must be a generic way of passing the participant options in order to have a generic test. So we either need a generic builder or ParticipantOptions interface to be passed by the test. The ParticipantOptions describe what kind of participant will be created: is it chrome, firefox, android, ios ? is it local or does it run on selenium grid etc.

Isn't that already there?  In `PariticipantFactory` it takes in `ParticipantOptions` which has a type.

But to take a step back (because I think I'm getting lost a bit).  Do these values come from a config file or something?  If so then I'll shut up because the props thing is probably fine...a builder may not make much sense there (depending on the nature of the properties, I guess).  So is that also what this change is?  Changing the mobile participant configuration from being located in the code to being located in a properties file?

paweldomas:
Yes, the properties can come from a config file (settings.properties) or java's command line -D argument.

paweldomas:
Mobile configuration previously was also fed from .properties file, but the code was providing defaults (and it still is).

paweldomas:
>Isn't that already there? In ParticipantFactory it takes in ParticipantOptions which has a type.
So at that point we don't know yet what participant's type will be. It could be that this prefix has different participant type defined in the properties file.

src/test/java/org/jitsi/meet/test/web/WebParticipantOptions.java
```diff
@@ -42,10 +50,80 @@
...
+     * loading chrome remotely like audio/video files.
+     */
+    private static final String PROP_REMOTE_RESOURCE_PARENT_PATH_NAME
+        = "remote.resource.path";
+
+    /**
+     * The property to disable no-sandbox parameter for chrome.
+     */
+    private static final String PROP_DISABLE_NOSANBOX
+        = "chrome.disable.nosanbox";
```
bgrozev: I'm guessing the typo ("sanbox") has been there all along?

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
```diff
@@ -221,13 +197,16 @@ else if (participantType == ParticipantType.edge)
 
             ops.setCapability(CapabilityType.LOGGING_PREFS, logPrefs);
 
-            if (!Boolean.getBoolean(PROP_DISABLE_NOSANBOX))
+            // FIXME This is hard to understand. The property is "to disable",
```
bgrozev: I completely agree with the comment. But why are we adding confusing method names on top of the confusing property name? We can name the method e.g. `sandboxEnabled` and implement the logic there.
And longer term we can deprecate the old property, add a new one, and handle both in `sandboxEnabled`.

bgrozev: I'm looking at the PR one commit at a time, so I might have missed changes in later commits.

src/test/java/org/jitsi/meet/test/web/WebParticipantOptions.java
```diff
@@ -55,11 +133,63 @@ protected Properties initDefaults()
...
+    {
+        return getBooleanProperty(PROP_DISABLE_NOSANBOX);
+    }
+
+    /**
+     * The getter for {@link #PROP_ENABLE_HEADLESS}.
+     *
+     * @return <tt>true</tt> or <tt>false</tt>.
+     */
+    public boolean isEnabledHeadless()
```
bgrozev: headlessEnabled? isHeadless?

src/test/java/org/jitsi/meet/test/mobile/base/AndroidParticipantOptions.java
```diff
@@ -13,24 +13,24 @@
...
 
-import org.jitsi.meet.test.base.*;
+import org.openqa.selenium.remote.*;
 
 import java.util.*;
 
 /**
- * The Android version of {@link MobileParticipantBuilder}. Holds all the stuff
- * specific to Android {@link MobileParticipant}.
+ * The set of options which configures an iOS
```
bgrozev: s/iOS/Android/?

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
damencho: So the option we pass to chrome is 'no-sandbox', and before that the static variable was DISABLE_NOSANBOX_PARAM, which is better when reading than PROP_DISABLE_NOSANBOX, but this can be only me. I know that probably it became PROP_DISABLE_NOSANBOX to keep consistency with other props.

src/test/java/org/jitsi/meet/test/mobile/base/MobileParticipantOptions.java
```diff
@@ -15,13 +15,215 @@
...
+            = Objects.requireNonNull(
+                    getAppiumServerAddress(), "appiumServerAddress");
+        String appiumPort
+            = Objects.requireNonNull(
+                    getAppiumServerPort(), "appiumServerPort");
+
+        try
+        {
+            return new URL(
+                "http://" + appiumAddress + ":" + appiumPort + "/wd/hub");
```
bgrozev: is http instead of https on purpose?

src/test/java/org/jitsi/meet/test/mobile/base/iOSParticipantOptions.java
```diff
@@ -13,19 +13,20 @@
...
 
-import org.jitsi.meet.test.base.*;
+import org.openqa.selenium.remote.*;
 
 import java.util.*;
 
 /**
- * The iOS version of {@link MobileParticipantBuilder} which holds all the iOS
- * specific stuff.
+ * The set of options which configures an Android
```
bgrozev: s/Android/iOS/? :)

src/test/java/org/jitsi/meet/test/mobile/base/MobileParticipantFactory.java
```diff
@@ -47,8 +47,8 @@ public MobileParticipantFactory(Properties config)
      * {@inheritDoc}
      */
     @Override
-    public Participant<? extends WebDriver> createParticipant(
-            String configPrefix, ParticipantOptions options)
+    public Participant<? extends WebDriver> doCreateParticipant(
```
bgrozev: Does this need to stay public? The overriden method is protected

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
```diff
@@ -59,9 +59,11 @@ public WebParticipantFactory(Properties config)
         super(WebParticipantOptions.moveLegacyGlobalProperties(config));
     }
 
+    /**
+     * {@inheritDoc}
+     */
     @Override
-    public Participant<? extends WebDriver> createParticipant(
-        String configPrefix,
+    public Participant<? extends WebDriver> doCreateParticipant(
```
bgrozev: Same as above

src/test/java/org/jitsi/meet/test/mobile/base/MobileParticipantOptions.java
damencho: Yep, connections to the selenium grid are http.

damencho:
So, I'm ok with the changes. While I was working on adding ParticipantOptions I made several proposals for that change. And we were discussing those with Pawel to come to the current situation. I think one of my proposals was something similar to this PR with abstract ParticipantFactory (but without refactoring the mobile part, I think). 

In the end, all the stuff under *.base, *.mobile, and *.web are hidden by devs implementing tests, so it just needs to work and to be easy to navigate and read :) And this packages once we stop refactoring should not be changed frequently, except some props that we pass to browsers or some more debug stuff in failureListener.

Yep, it is not a trivial thing to merge both web and mobile. But to my point, I really want to agree on something, cause we saw not a few changes in these packages and I prefer to spend time writing and running more tests, as this is what we need and leave some stuff for future improvements :)
At the end these are tests, to have some test with not perfect code, is better than having none. Tests are supposed to be implemented quickly, with not spending a lot of time on optimization or quality, but to make them work. So I think we are in a good shape, we have already the tools for that with the abstract test classes, with participants, with participantOptions ... so it is relatively easy and quick to write test.

damencho:
And just to say that I didn't have time to review it in details as Bobi is doing :P Thanks for stepping in.

src/test/java/org/jitsi/meet/test/mobile/base/MobileParticipantFactory.java
paweldomas: right, that's by mistake

src/test/java/org/jitsi/meet/test/mobile/base/MobileParticipantOptions.java
paweldomas: I'm not sure if https will work, but will try that and change to http or even go with the full address if possible, but I think it was split into address and port because of Amazon device farm which provides those 2 in system properties

src/test/java/org/jitsi/meet/test/web/WebParticipantOptions.java
paweldomas: thanks

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
bgrozev: I think this is easier to read
```if (!options.enableSandbox())
{
    ops.addArguments("no-sandbox");
}```

Compared to `if (!options.isDisableNoSandbox())`

Or better: `if(options.disableSandbox())`

**Comment by bgrozev** [comment]
> And just to say that I didn't have time to review it in details as Bobi is doing

And I could only review the details, I don't understand the overall architecture :)

File: src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java (hunk 2 above)
Comment (paweldomas) [reply src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java]: I will go with "disableSandbox" which was your last "or better" comment assuming it's the preferred way.

File: src/test/java/org/jitsi/meet/test/web/WebParticipantOptions.java (hunk 1 above)
Comment (paweldomas) [reply src/test/java/org/jitsi/meet/test/web/WebParticipantOptions.java]: Yes, it was there and I will be keeping it as is with a comment. Probably it would not be a good idea to rename it now given that it will be deprecated by the new property name discussed below.

**Comment by paweldomas** [comment]
Hey @bgrozev I have addressed your comments. Please take a look and let me know how it looks now. Upon approval I will squash some of the commits as I would rather keep the commits separate in the end.

**Comment by bgrozev** [comment]
> At the end these are tests, to have some test with not perfect code, is better than having none. Tests are supposed to be implemented quickly, with not spending a lot of time on optimization or quality, but to make them work.

I agree, but if we want to be able to write tests quickly, we need to have a good base. I have spent a lot of time trying to understand how some hastily written piece of code works so that I can adapt it to my case.
And right now we are in a MUCH better situation than before, but that doesn't mean we can stop worrying about the code quality and start writing tests. Let's keep the quality high.

**Comment by damencho** [comment]
Of course we are in better shape than before, I don't want to affect quality in any way, but I'm worried that we are at the point where we have done a lot and situation in test classes seems fine, and any further refactoring in those base classes will start to seem over-engineering and waste of time we need that time to be writing tests. 
We also need to spend time and write and run some hybrid tests to see how it goes ... there can be some stuff/problems to surface while adding tests a d running them, we need some time in practice... Probably there are more stuff to fix and make better but let's see what are those ...

File: src/test/java/org/jitsi/meet/test/mobile/base/MobileParticipantFactory.java [hunk 9]
```diff
@@ -38,21 +43,71 @@ public MobileParticipantFactory(Properties config)
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     @Override
-    public Participant<? extends WebDriver> createParticipant(
-        String configPrefix,
-        ParticipantOptions options)
+    public Participant<? extends WebDriver> doCreateParticipant(
```
bgrozev: And this is why I don't like force pushing :)

paweldomas: what about this one ?

bgrozev: To clarify: this was fixed to "protected" in the previous version of the PR (6d7b4a74e6e8d0a5f30809b3b56bd03b6b800cdb), and it's not easy to spot the difference in the gh interface.

paweldomas: ok, got it

paweldomas: good job on catching this one :)

PR #156, opened 2018-03-27
feat: Adds more networks (and a minor fix)

gpolitis:
Please do NOT merge this yet.


gpolitis:
@bbaldino could you PTAL?

gpolitis:
With a relay candidate the port corresponds to the port allocated that the TURN server has allocated for the receiver, but the TURN server would send media to the receiver on a different port, invisible to the receiver. For now this seems to work but please let me know if you think we can improve this in any way.

PR #157, opened 2018-03-28
ref(toolbar): convert tests to use the toolbar page object

virtuacoplenny:


virtuacoplenny:
Okay, I was gonna do a bit more cleanup tomorrow morning. I overrode some methods while moving stuff over and the old methods can be removed. Let me know which import needs to be changed, or anything else that needs changing, and I'll do it.

virtuacoplenny:
I'll probably also rename the toolbar methods to be getXXButton to be consistent with the methods that already exist.

src/test/java/org/jitsi/meet/test/util/MeetUIUtils.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.jitsi.meet.test.util;
 
+import org.jitsi.meet.test.web.WebParticipant;
```
damencho: This one.

PR #158, opened 2018-03-29
Fix audioonly clients

damencho:
I had separated the changes in different commits. If you prefer I can do separate PRs.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/StartMutedTest.java
+++ b/src/test/java/org/jitsi/meet/test/StartMutedTest.java
@@ -57,7 +57,13 @@ public class StartMutedTest
         driver1.findElement(By.id("startAudioMuted")).click();
         driver1.findElement(By.id("startVideoMuted")).click();
 
-        ensureTwoParticipants();
+        WebParticipant participant2 = joinSecondParticipant();
+
+        // if the participant2 is audio only and audio muted there will be no
+        // data sent
+        participant2.waitForSendReceiveData(
+            !participant2.isAudioOnlyParticipant(),
+            true);
 
         // On the PR testing machine it seems that some audio is leaking before
         // we mute. The audio is muted when 'session-initiate' is received, but
@@ -76,19 +82,24 @@ public class StartMutedTest
     {
         hangUpAllParticipants();
 
-        ensureTwoParticipants(
-            getJitsiMeetUrl().appendConfig(
-                "config.startAudioMuted=1&" +
+        ensureOneParticipant(getJitsiMeetUrl().appendConfig(
+            "config.startAudioMuted=1&" +
                 "config.debugAudioLevels=true&" +
-                "config.startVideoMuted=1"),
-            null);
+                "config.startVideoMuted=1"));
+
+        WebParticipant participant2 = joinSecondParticipant();
+
+        // if the participant is audio only, if audio is muted we will not
+        // receive any data, so skip upload check
+        participant2.waitForSendReceiveData(
+            !participant2.isAudioOnlyParticipant(), true);
 
         WebParticipant participant1 = getParticipant1();
 
-        final WebDriver driver2 = getParticipant2().getDriver();
+        final WebDriver driver2 = participant2.getDriver();
         participant1.executeScript(
             "console.log('Start configOptionsTest, second participant: "
-                + getParticipant2().getEndpointId() + "');");
+                + participant2.getEndpointId() + "');");
 
         participant1.waitForIceConnected();
 
@@ -101,7 +112,7 @@ public class StartMutedTest
         checkParticipant2ForMute();
 
         // Unmute and see if the audio works
-        getParticipant2().getToolbar().clickAudioMuteButton();
+        participant2.getToolbar().clickAudioMuteButton();
         participant1.executeScript(
             "console.log('configOptionsTest, unmuted second participant');");
         MeetUIUtils.waitForAudioMuted(
--- a/src/test/java/org/jitsi/meet/test/StartMutedTest.java
+++ b/src/test/java/org/jitsi/meet/test/StartMutedTest.java
@@ -117,40 +117,21 @@ public class StartMutedTest
      */
     private void checkParticipant2ForMute()
     {
-        WebDriver driver1 = getParticipant1().getDriver();
-        WebDriver driver2 = getParticipant2().getDriver();
+        WebParticipant participant1 = getParticipant1();
+        WebParticipant participant2 = getParticipant2();
 
-        TestUtils.waitForElementByXPath(
-            driver2,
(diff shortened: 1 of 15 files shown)


src/test/java/org/jitsi/meet/test/web/WebParticipant.java
```diff
@@ -296,6 +297,56 @@ public boolean isXmppConnected()
                 "return APP.conference._room.xmpp.connection.connected;"));
     }
 
+    /**
+     * {@inheritDoc}
+     * If it is not audio muted will be checking it for 5 seconds before
+     * returning a result.
+     */
+    @Override
+    public boolean isAudioMuted()
```
paweldomas: If the implementation is waiting then for me it would make sense to change isAudioMuted to assertAudioMuted(boolean). Same for the video one.

"Participant2 is not audio muted" this text can be generated on assertFailure by the participant which then can use it's name.

src/test/java/org/jitsi/meet/test/pageobjects/web/Filmstrip.java
```diff
@@ -0,0 +1,88 @@
...
+     * Asserts that {@code participant} shows or doesn't show the audio
+     * mute icon for the conference participant identified by
+     * {@code participantToCheck}.
+     *
+     * @param participantToCheck the {@code WebParticipant} for whom we're
+     * checking the status of audio muted icon.
+     * @param isDisplayed if {@code true}, the method will assert the presence
+     * of the "mute" icon; otherwise, it will assert its absence.
+     */
+    public void assertAudioMuteIcon(
```
paweldomas: Can we make Filmstrip an interface and accept Participant instead of WebParticipant ? We have filmstrip on mobile as well, but for now getFilmstrip() would return null on mobile.

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
```diff
@@ -554,4 +555,17 @@ public Toolbar getToolbar()
 
         return toolbar;
     }
+
+    /**
+     * @return a representation of the filmstrip of this participant.
+     */
+    public Filmstrip getFilmstrip()
```
paweldomas: In the context of the comment above this would land in the Participant instead

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
damencho: This is just checking whether it is audioMuted or not. The wait is just giving time the presence to be received. I just used the method in some asserts. 
This case is covered in the last commit of the PR, which reminds me that I changed the timeout there and I should revert it back to 5.
StartMutedTest for example, participant 1 is muted, participant 2 is joining but because it start muted and it is Safari, not dealing with video, we should skip waitForSendReceiveData cause there will be none.

src/test/java/org/jitsi/meet/test/web/WebTestBase.java
```diff
@@ -174,13 +174,24 @@ private void ensureTwoParticipantsInternal(
...
         participant.waitToJoinMUC(10);
 
         participant.waitForIceConnected();
-        participant.waitForSendReceiveData();
+
+        // if the participant is audio only, if audio is muted we will not
+        // receive any data, so skip download check
+        // if the participant is audio only and audio muted there will be no
+        // data sent
+        participant.waitForSendReceiveData(
```
paweldomas: in which tests this is a problem ? I'm trying to think of better solution...

To me it seems it would be better if only the problematic test is making such decision, because it knows it's the case.

Wouldn't it be nicer if Safari WebParticipant, which is aware of it's limitations, if it would ignore the requirement to both send and receive at the same time ?

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
paweldomas: It feels weird that getter method is waiting for anything. It's implementation clearly looks like an assert.

src/test/java/org/jitsi/meet/test/web/WebTestBase.java
damencho: there are cases which need to check only sending, and cases where it can check only receiving and cases where it is fine to check both.
We can add parameters to ensureParticipant methods, or we can be passing it through ParticipantOptions? WDYT?

paweldomas: Then those special places can create participants manually, and not use the "ensure" method. They will know whether to call waitForSendReceive with true or false. The "ensure" method should be a convenience method which covers only standard case. Getting "firstParticipant" here just feels like a big complication.

damencho:
@paweldomas I did some modifications, as we talked offline. Can you take a look?

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
```diff
@@ -360,8 +371,31 @@ public void waitForSendReceiveData()
                     long download = bitrate.get("download");
                     long upload = bitrate.get("upload");
 
-                    if (download > 0 && upload > 0)
-                        return true;
+                    if (checkSend)
```
paweldomas: can this be simplified to this ?
`return (!checkSend || upload > 0) && (!checkReceive || download > 0);`

src/test/java/org/jitsi/meet/test/StartMutedTest.java
```diff
@@ -57,7 +57,16 @@ public void checkboxesTest()
         driver1.findElement(By.id("startAudioMuted")).click();
         driver1.findElement(By.id("startVideoMuted")).click();
 
-        ensureTwoParticipants();
+        WebParticipant participant = joinSecondParticipant();
```
paweldomas: it would be more readable if this var would be named 'participant2' - when reading this I was thinking it's "participant1"

src/test/java/org/jitsi/meet/test/StartMutedTest.java
```diff
@@ -57,7 +57,16 @@ public void checkboxesTest()
...
+
+        // if the participant is audio only, if audio is muted we will not
+        // receive any data, so skip download check
+        // if the participant is audio only and audio muted there will be no
+        // data sent
+        participant.waitForSendReceiveData(
+            !participant.isAudioOnlyParticipant(),
+            !participant.isAudioOnlyParticipant());
```
paweldomas: shouldn't it be:
`participant2.waitForSendReceiveData(`
`    !participant2.isAudioOnlyParticipant(),`
`    !participant1.isAudioOnlyParticipant());`
?

damencho: I think I had fixed that and force pushed the second commit of the PR.

PR #159, opened 2018-03-30
feat(toolbar): new page object for new toolbar

virtuacoplenny:
I've left in the old toolbar page object. I'll be getting rid of it when the old toolbar is removed from jitsi-meet. I suppose as long as the old toolbar exists I'd prefer to keep it present so it's testable.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/ContactListTest.java
+++ b/src/test/java/org/jitsi/meet/test/ContactListTest.java
@@ -21,6 +21,7 @@ import org.jitsi.meet.test.web.*;
 
 import org.openqa.selenium.*;
 import org.openqa.selenium.support.ui.*;
+import org.testng.*;
 import org.testng.annotations.*;
 
 import static org.testng.Assert.*;
@@ -53,6 +54,11 @@ public class ContactListTest
     public void testContactList()
     {
         WebParticipant participant1 = getParticipant1();
+
+        if (participant1.getToolbar().isNewToolbar()) {
+            throw new SkipException(
+                "Contact list no supported in new toolbox. Disabling test.");
+        }
 
         // Make sure that the contact list panel is open.
         MeetUIUtils.displayContactListPanel(participant1);
--- a/src/test/java/org/jitsi/meet/test/ContactListTest.java
+++ b/src/test/java/org/jitsi/meet/test/ContactListTest.java
@@ -57,7 +57,7 @@ public class ContactListTest
 
         if (participant1.getToolbar().isNewToolbar()) {
             throw new SkipException(
-                "Contact list no supported in new toolbox. Disabling test.");
+                "Contact list not supported in new toolbox. Disabling test.");
         }
 
         // Make sure that the contact list panel is open.
--- a/src/test/java/org/jitsi/meet/test/pageobjects/web/ToolbarV2.java
+++ b/src/test/java/org/jitsi/meet/test/pageobjects/web/ToolbarV2.java
@@ -1,0 +1,416 @@
+/*
+ * Copyright @ 2018 Atlassian Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.meet.test.pageobjects.web;
+
+import org.jitsi.meet.test.util.*;
+import org.jitsi.meet.test.web.*;
+import org.openqa.selenium.*;
+
+import java.util.*;
+
+/**
+ * Represents the toolbar in a particular {@link WebParticipant}.
+ *
+ * @author Leonard Kim
+ */
+public class ToolbarV2 extends Toolbar {
+    /**
+     * Accessibility labels to be used as selectors for finding WebElements
+     * within the {@link ToolbarV2}.
+     */
+    private final static String AUDIO_MUTE = "Audio mute";
+    private final static String CHAT = "Chat";
+    private final static String DESKTOP = "Screenshare";
+    private final static String ETHERPAD = "Etherpad";
+    private final static String HANGUP = "Hangup";
+    private final static String INFO = "Info";
+    private final static String OVERFLOW = "Overflow";
+    private final static String OVERFLOW_MENU = "Overflow menu";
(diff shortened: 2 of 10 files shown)


src/test/java/org/jitsi/meet/test/pageobjects/web/ToolbarV2.java
```diff
@@ -0,0 +1,416 @@
...
+import org.openqa.selenium.*;
+
+import java.util.*;
+
+/**
+ * Represents the toolbar in a particular {@link WebParticipant}.
+ *
+ * @author Leonard Kim
+ */
+public class ToolbarV2 extends Toolbar {
```
virtuacoplenny: I inherited out of laziness to inherit the type. The old toolbar should be removed soon-ish. I'm eyeballing a month from now unless someone has other plans. I'm not sure if this is an acceptable short term hack.

src/test/java/org/jitsi/meet/test/ContactListTest.java
```diff
@@ -54,6 +55,11 @@ public void testContactList()
     {
         WebParticipant participant1 = getParticipant1();
 
+        if (participant1.getToolbar().isNewToolbar()) {
+            throw new SkipException(
+                "Contact list no supported in new toolbox. Disabling test.");
```
bbaldino: `no` -> `not`

src/test/java/org/jitsi/meet/test/pageobjects/web/ToolbarV2.java
```diff
@@ -0,0 +1,416 @@
...
+    /**
+     * Accessibility labels to be used as selectors for finding WebElements
+     * within the {@link ToolbarV2}.
```
bbaldino: would it make sense to add a reference in this comment to the js file in which the values these are mapped to are located?

src/test/java/org/jitsi/meet/test/pageobjects/web/ToolbarV2.java
```diff
@@ -0,0 +1,416 @@
...
+     * Checks whether or not the recording button is present in the toolbar.
+     *
+     * @return True if the recording button is present, false if it is not.
+     */
+    public boolean hasRecordButton()
+    {
+        openOverflowMenu();
+
+        List<WebElement> elements
+            = participant.getDriver().findElements(By.cssSelector(
```
bbaldino: this is the only `hasXXXButton` method i assume? that's why there isn't a generic `hasButton(String label)` method?

src/test/java/org/jitsi/meet/test/pageobjects/web/ToolbarV2.java
```diff
@@ -0,0 +1,416 @@
...
+     * clicked.
+     */
+    private void waitForButtonDisplay(String accessibilityLabel)
+    {
+        TestUtils.waitForElementBy(
+            participant.getDriver(),
+            By.cssSelector(getAccessibilityCSSSelector(accessibilityLabel)),
+            10);
+    }
+}
```
bbaldino: nice helpers in here, keeps a lot of much out of the way when you're reading the methods.

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
```diff
@@ -498,7 +498,14 @@ public Toolbar getToolbar()
     {
         if (toolbar == null)
         {
-            toolbar = new Toolbar(this);
+            boolean usingNewToolbar = (boolean) executeScript
+                ("return interfaceConfig._USE_NEW_TOOLBOX");
```
bbaldino: i don't think we ever break lines like this in java.  at least put the opening brace of the method call next to the `executeScript`:
```
boolean usingNewToolbar = (boolean)executeScript(
    "return interfaceConfig._USE_NEW_TOOLBOX");
```
but i think even better (if it fits) would be:
```
boolean usingNewToolbar = 
    (boolean)executeScript("return interfaceConfig._USE_NEW_TOOLBOX");
```

src/test/java/org/jitsi/meet/test/pageobjects/web/ToolbarV2.java
virtuacoplenny: I'm not sure. I see that as potentially a bug because the file can change and someone wouldn't know to update this comment.

src/test/java/org/jitsi/meet/test/pageobjects/web/ToolbarV2.java
virtuacoplenny: Hm, very likely but not intentionally. I was breaking out helpers when I saw some re-use potential, and it looks like only JibriTest uses a hasXXXButton method. It seems like all the other tests assume the button will be present.

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
virtuacoplenny: I went with the first, line breaking after executeScript, due the second exceeding the 80 character line limit.

paweldomas: hmm I think we used to put the = sign on the next line (at least I do that not sure if it's part of the official style guidelines)

src/test/java/org/jitsi/meet/test/pageobjects/web/ToolbarV2.java
virtuacoplenny: Could the simliar not be said for all selectors that have been used?

virtuacoplenny: My last comments actually doesn't address your comment at all. It provides something loosely connected to the issue of adding comments in jitsi-meet saying "this is used for torture." Let's say I do add the comments to jitsi-meet. I have declared aria-label in 5 separate files. Would you then want me to add a comment to each? An alternative would be to group the labels into a constants and add one comment there. Or make the assumption that people know aria-labels are used for both torture and screen readers.

bbaldino: > Could the simliar not be said for all selectors that have been used?

Yeah I guess so


> Let's say I do add the comments to jitsi-meet. I have declared aria-label in 5 separate files. Would you then want me to add a comment to each? An alternative would be to group the labels into a constants and add one comment there. Or make the assumption that people know aria-labels are used for both torture and screen readers.

I guess I was just worried about the potential for updating one and forgetting to update another, but I see what you mean about it being sticky and there being lots of places to update.  Considering the tests will fail in this scenario anyway (and I think in a reasonably obvious way?) then I guess it feels ok as is.

virtuacoplenny: Maybe what you're liking is the idea of something like a data-qa attribute that's explicit. data-qa isn't some standard attribute, it stands out a bit, and it would have one use: to be used as selectors for torture tests. How this could apply to mobile I wouldn't know, but ignoring mobile I think that would address your concerns. Would you prefer I added data-qa (or a similar name) in additional to aria-label?

PR #160, opened 2018-03-30
Make it possible to enable CallStats in the tests.

gpolitis:


gpolitis:
@damencho could you PTAL?

PR #161, opened 2018-03-30
Fix join conference Android 8

paweldomas:
On newer Android phones the old UI automator does not find the "conference" element after entering the room. See commit msg for more details.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/mobile/MobileParticipant.java
+++ b/src/test/java/org/jitsi/meet/test/mobile/MobileParticipant.java
@@ -249,9 +249,16 @@ public class MobileParticipant extends Participant<AppiumDriver<WebElement>>
 
         if (type.isAndroid())
         {
-            driver.hideKeyboard();
 
-            welcomePageView.getJoinRoomButton().click();
+            // FIXME The 3 clicks below is a hack to workaround a bug in Jitsi
+            // Meet UI where the focus is lost after the first click.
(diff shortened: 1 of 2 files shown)


src/test/java/org/jitsi/meet/test/mobile/MobileParticipant.java
```diff
@@ -249,9 +249,16 @@ public void doJoinConference(JitsiMeetUrl conferenceUrl)
...
 
-            welcomePageView.getJoinRoomButton().click();
+            // FIXME The 3 clicks below is a hack to workaround a bug in Jitsi
+            // Meet UI where the focus is lost after the first click.
+            // To be removed once the bug mentioned above is fixed (I don't know
+            // how to fix it).
+            roomNameInput.click();
+            roomNameInput.click();
+            roomNameInput.click();
+
```
virtuacoplenny: Do you know why the three clicks works?

virtuacoplenny: I guess a follow up question would be: is the bug really so complicated that it's worth working around instead of addressing for now?

paweldomas: We have a ticket for the issue, so it will be worked on. The problem is that mobile tests will not work at all on some devices without the workaround. I tried looking into that code, but I don't know how to fix it. Probably @zbettenbuk who wrote it will know better.

PR #162, opened 2018-04-02
WIP: Compare Kalman filter vs linear regression GCC

gpolitis:
please do NOT merge this yet

### DIFF
--- a/src/test/java/org/jitsi/meet/test/base/AbstractTest.java
+++ b/src/test/java/org/jitsi/meet/test/base/AbstractTest.java
@@ -205,15 +205,6 @@ public abstract class AbstractTest
     }
 
     /**
-     * Tests can override this to skip saving debug information (such as
-     * RTP stats, HTML source code, screenshots etc.)
-     */
-    public boolean saveDebugInformation()
-    {
-        return true;
-    }
-
-    /**
      * Tests can override this to be disabled by default and run only when
      * explicitly included in the tests to run using
      * {@link AbstractBaseTest#TESTS_TO_RUN_PNAME} or
--- a/src/test/java/org/jitsi/meet/test/base/AbstractTest.java
+++ b/src/test/java/org/jitsi/meet/test/base/AbstractTest.java
@@ -1,0 +1,263 @@
+/*
+ * Copyright @ 2015 Atlassian Pty Ltd
+ *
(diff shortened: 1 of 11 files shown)


src/test/java/org/jitsi/meet/test/base/AbstractTest.java
```diff
@@ -0,0 +1,263 @@
...
+
+/**
+ * The base of all tests, checks whether the tests is supposed to be executed
+ * or skipped by checking the properties which tests to run, exclude or include.
+ *
+ * @author Damyan Minkov
+ * @author George Politis
+ */
+@Listeners(FailureListener.class)
+public abstract class AbstractTest
```
paweldomas: so will we have AbstractTest and AbstractBaseTest classes ? can we come up with better naming, so that it would indicate what's the difference between the two ? Otherwise we could as well name them AbstractBaseTest1 and AbstractBaseTest2...

gpolitis: Thanks for the comments Pawel, I'll try to come up with a better name for the new class.

The bwe estimation tests do need participant management (if you've spotted a place where it says differently please point that out to me and I'll fix it), but the lifecycle is different from what we have now: In the bwe estimation tests the test method is executed in parallel and each parallel test method launches 2 participants. Does that shed some light?

paweldomas: Yes, it does shed some lights. I don't like how it duplicates the lifecycle procedure. There must be another way to make it parallel. I'll try to think about it.

paweldomas: So one way would be to compose a class which would extend WebTestBase. Then inside the "test"/"runScenario" method you can manually set it up by passing the Properties to it's "setupClassPrivate". After the test is done just call it's cleanup methods.

Because of that duplication you have already skipped one step from the initialisation phase:
https://github.com/jitsi/jitsi-meet-torture/blob/master/src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java#L133

gpolitis:
Thanks for your comments @paweldomas and for the discussion that we had offline. I will open a new PR that follows your recommendations.

PR #176, opened 2018-04-21
Add timeout for persitency check before reload

zbettenbuk:
This test randomly fails since the redux-based settings support on web. I think the reason is that the persistency layer has a 2 seconds throttling to avid overuse of localStorage and sometimes it just doesn't always get to run before the test browser is closed/reloaded.


virtuacoplenny:
The current test failure in this PR is because the dang info dialog is covering a remote thumbnail. Let me look at that again...

damencho:
Merged https://github.com/jitsi/jitsi-meet-torture/pull/179

PR #178, opened 2018-04-24
Debug logs

damencho:
Depends on https://github.com/jitsi/jitsi-meet/pull/2858

### DIFF
--- a/src/test/java/org/jitsi/meet/test/base/Participant.java
+++ b/src/test/java/org/jitsi/meet/test/base/Participant.java
@@ -490,17 +490,7 @@ public abstract class Participant<T extends WebDriver>
      */
     public abstract String getRTPStats();
 
-    public LogEntries getBrowserLogs()
-    {
-        if (type.isFirefox())
-        {
-            // not currently supported in FF
-            // https://github.com/SeleniumHQ/selenium/issues/2910
-            return null;
-        }
-
-        return driver.manage().logs().get(LogType.BROWSER);
-    }
+    public abstract List getBrowserLogs();
 
     /**
      * Returns the value for the given <tt>key</tt> from the config.js loaded
--- a/src/test/java/org/jitsi/meet/test/web/WebParticipant.java
+++ b/src/test/java/org/jitsi/meet/test/web/WebParticipant.java
@@ -25,6 +25,7 @@ import org.openqa.selenium.interactions.*;
 import org.openqa.selenium.remote.*;
 import org.openqa.selenium.support.ui.*;
 
+import java.util.*;
 import java.util.concurrent.*;
 import java.util.logging.*;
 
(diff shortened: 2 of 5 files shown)


src/test/java/org/jitsi/meet/test/web/WebParticipant.java
```diff
@@ -514,4 +515,10 @@ public WebFilmstrip getFilmstrip()
 
         return filmstrip;
     }
+
+    @Override
+    public List getBrowserLogs()
+    {
+        return (List)executeScript("return APP.debugLogs.getLogs();");
```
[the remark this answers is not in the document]
paweldomas: also there's missing space after (List)

src/test/java/org/jitsi/meet/test/base/Participant.java
```diff
@@ -490,17 +490,7 @@ public void saveHtmlSource(File outputDir, String fileName)
...
-        if (type.isFirefox())
-        {
-            // not currently supported in FF
-            // https://github.com/SeleniumHQ/selenium/issues/2910
-            return null;
-        }
-
-        return driver.manage().logs().get(LogType.BROWSER);
-    }
+    public abstract List getBrowserLogs();
```
paweldomas: can we specify what kind of list that is ? List<Object> or whatever

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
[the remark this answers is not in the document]
damencho: This is fixed.

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
```diff
@@ -514,4 +516,21 @@ public WebFilmstrip getFilmstrip()
...
     }
+
+    @Override
+    public List getBrowserLogs()
+    {
+        try
+        {
+            return (List) executeScript("return APP.debugLogs.getLogs();");
+        }
+        catch (Throwable t)
```
paweldomas: not Throwable, Exception at best, but ideally a RuntimeException

src/test/java/org/jitsi/meet/test/base/Participant.java
paweldomas: sorry, I meant `List<Object>`

damencho: What is the point of having that?

paweldomas: Because in one doc you write it's a string in other there can be just anything and it's not obvious what's there. It's just kind of self documenting, but you can leave it as is if you don't like that

paweldomas: yeah, maybe I just got used to templates somehow

PR #181, opened 2018-05-02
Retry joining a muc on first failure.

damencho:
With latest chrome update (v66, v65) we see sometimes on screenshots and on video that the browser is stuck on the permissions screen and never joins the room. We skip the first wait to join failure and we retry second time.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/web/WebTestBase.java
+++ b/src/test/java/org/jitsi/meet/test/web/WebTestBase.java
@@ -368,12 +368,20 @@ public class WebTestBase
     {
         WebParticipant participant = joinParticipant(index, meetURL, options);
 
+        // FIXME: remove this when chrome changes and we stop seeing the warning
+        // in the tests logs
         try
         {
             participant.waitToJoinMUC(10);
         }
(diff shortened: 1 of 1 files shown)


bbaldino:
do we know what causes the failure to join?

damencho:
No idea. I have console logs and video. Seems to me gum never succeed or fail, the promise is never resolved or rejected. At least that what it looks like from the console logs.

damencho:
Jenkins, test this please.

src/test/java/org/jitsi/meet/test/web/WebTestBase.java
```diff
@@ -368,6 +368,16 @@ private WebParticipant joinParticipantAndWait(
     {
         WebParticipant participant = joinParticipant(index, meetURL, options);
 
+        // the try catch below is only chrome workaround
+        if (!participant.getType().isChrome())
+        {
```
paweldomas: will it be less code if in the try catch below there will be a check at the top: if(!participant.getType().isChrome()) throw ex; ?

PR #187, opened 2018-05-31
chore(i18n): update accessibility labels

guusdk:
This brings the value of accessibility labels in line with the changes introduced
by https://github.com/jitsi/jitsi-meet/pull/3071


damencho:
Ok, works now. It is just the ios failure, it has something to do with the device, I tried fixing it, but it still doesn't work will take a look later today.

guusdk:
Ah, good. You reviewed the new test results before I remembered I had to :)
Does this also mean it's OK to merge https://github.com/jitsi/jitsi-meet/pull/3071 - or does that require more review?

damencho:
I just want to sync up with @virtuacoplenny and then it is ok to merge.

PR #189, opened 2018-06-12
improvement: allow jitsi-meet instance URL to have a trailing slash.

guusdk:
To configure against what server these tests are executed, a URL is provided
through system properties, like this:

  mvn test -Djitsi-meet.instance.url="https://meet.example.com"

This commit prevents failures when this URL ends with a slash. In other words,
this is now also acceptable:

  mvn test -Djitsi-meet.instance.url="https://meet.example.com/"


guusdk:
I ran into this, as I'm hosting meet on something different than the root context. The errors that followed confused me for a bit, which this change will prevent from happening again.

PR #190, opened 2018-06-13
fix(mobile): set implicit wait for variable display timings

virtuacoplenny:
~~I don't know how to run the tests locally so I'm just gonna run them here...~~

The current tests do not set an implicit wait timeout. The tests assume then that overlays and dialogs will display immediately. It could be that there is a display delay, so actions to dismiss a dialog or overlay doesn't happen at the right time and blocks further test progress.

That's the guess...


virtuacoplenny:
Jenkins test this please.

virtuacoplenny:
Jenkins test this please.

PR #192, opened 2018-06-15
[RN] Allow calendar permission on start

zbettenbuk:
Since the calendar tab changes we now have 2 permission popups on start so we need to click allow for both.

Required by https://github.com/jitsi/jitsi-meet/pull/3095 which defaults to the Calendar tab and thus may cause the appearance of the Calendar permission prompt on startup.


lyubomir:
@zbettenbuk, in the future, please make sure to explicitly "inter-mention" related PRs in the PRs' descriptions so that reviewers have as much context as possible and don't miss interdependent PRs. We recently got in a situation where an author had two interdependent PRs but the reviewer was told of and merged one of them only, and then we had to involve a third-developer to merge the second PR after rebasing and new testing.

PR #196, opened 2018-07-06
feat(iOS): adds support for automatic WDA signing

paweldomas:
Adds support for the automatic WebDriverAgent Runner's signing described here:

https://github.com/appium/appium-xcuitest-driver/blob/master/docs/real-device-config.md#basic-automatic-configuration


paweldomas:
@virtuacoplenny can you land it ? it seems that it can be a different issue with the PR tests that the signing. I've checked on the machine and signing is configured correctly. I'll keep digging, but this PR will not hurt if lands - it will still be useful in future.

PR #199, opened 2018-07-13
fix(MobileParticipant): fix permissions on the welcome page

saghul:
No permissions are requested any longer. Camera and mic permissions are
now requested when joining.


saghul:
Required for https://github.com/jitsi/jitsi-meet/pull/3248

PR #200, opened 2018-07-13
fix(MobileParticipant): fix permissions on the welcome page

saghul:
No permissions are requested any longer. Camera and mic permissions are
now requeted when joining.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/mobile/MobileParticipant.java
+++ b/src/test/java/org/jitsi/meet/test/mobile/MobileParticipant.java
@@ -270,11 +270,7 @@ public class MobileParticipant extends Participant<AppiumDriver<MobileElement>>
             maybeAcceptOverlayPermissions();
         }
 
-        // ALLOW to use camera and the calendar
-        // Due to the async nature of Redux we don't know which permission will
-        // pop up first.
-        acceptPermissionAlert();
-        acceptPermissionAlert();
+        // XXX: No need to accept permissions on the welcome page now.
 
         WelcomePageView welcomePageView = new WelcomePageView(this);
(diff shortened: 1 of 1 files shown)


paweldomas:
The tests have failed because the iOS build got stuck. Not sure why. I'm giving it another try:
https://travis-ci.org/jitsi/jitsi-meet/builds

paweldomas:
Download and extract the tests-report.zip from here: http://ci.jitsi.org/view/pr-testing/job/jitsi-meet-torture-PR-test/219/
I'll write a manual on how to run mobile tests locally today

paweldomas:
It looks like it didn't accept the camera permissions for some reason. Lets see what happens after one more try.

saghul:
> I'll write a manual on how to run mobile tests locally today

Thank you!



> It looks like it didn't accept the camera permissions for some reason. Lets see what happens after one more try.

Hum, the whole point of this PR is that no camera permissions are requested :-) I guess it will run with the jitsi-meet build with the same branch name, right?

paweldomas:
Yes, it's supposed to run with the same branch name. After restart it failed with the 64 error code, but now Boris dismissed the dialog and I'm giving it another try.

saghul:
> On the screenshot it looks like it asks for camera permissions. So you're saying that it shouldn't ask ?

Correct. Which is what I observe when I test manually.

> I wonder if something's broken with branching and maybe the PR is not taking the jitsi-meet branch correctly or something.

This is my suspicion. Somehow Android works though.

saghul:
No problem @paweldomas! Thanks a lot for fixing that! ❤️

PR #206, opened 2018-09-05
feat(deps): update SettingsDialog page object for Tabs update

virtuacoplenny:


src/test/java/org/jitsi/meet/test/pageobjects/web/SettingsDialog.java
```diff
@@ -9,25 +9,31 @@
...
     private final static String START_VIDEO_MUTED_CHECKBOX
         = "[name='start-video-muted'] ~ div";
-    
+
+    /**
+     * XPath selectors to be used for finding tab label elements within the
+     * {@link SettingsDialog}. XPath is used so tabs can be found based on tab
+     * text, which CSS selectors cannot do.
+     */
+    private final static String MORE_TAB
```
paweldomas: as you can see above those constants used to be sorted alphabetically, so please lets stick to that

virtuacoplenny: I split them on purpose to separate CSS selectors from Xpath selectors. Should they still be merged into one group?

paweldomas: I'd say yes, because the IDE would not be able to figure this out. If you want to keep them together maybe add a prefix

PR #209, opened 2018-10-09
RN 57 related stuff

paweldomas:
Required for https://github.com/jitsi/jitsi-meet/pull/3508


paweldomas:
@saghul it seems to be fine landing it before the other things, so I'm merging

PR #211, opened 2018-11-03
Refactor some tests to use page objects

virtuacoplenny:


src/test/java/org/jitsi/meet/test/pageobjects/web/WebFilmstrip.java
```diff
@@ -66,6 +75,22 @@ public void assertAudioMuteIcon(
         );
     }
 
+    /**
+     * Asserts that the number of small videos with the dominant speaker
+     * indicator displayed equals 1.
+     */
+    public void assertOneDominantSpeaker()
```
damencho: The method doesn't check isDisplayed(), I just wonder is that check needed at all...

virtuacoplenny: I'm going to add it back. I don't think it's strictly needed but is still be better to keep because it's an explicit check for visibility. I think I had removed it because it seemed like old logic. Over a year ago in jitsi-meet the dominant speaker icons were always in the DOM, so it was important to check display to find the active dominant speaker. With reactification of the icons, only one dominant speaker icon should be in the DOM if things aren't broken.

virtuacoplenny:
Force pushed to resolve conflicts in TileViewTest.java.

virtuacoplenny:
Force pushed to rebase and fix conflicts in EtherpadTest and LockRoomTest.

virtuacoplenny:
I'm not going to keep fixing conflicts.

PR #216, opened 2018-11-30
feat(tile-view): double click to pin

virtuacoplenny:


virtuacoplenny:
@damencho, I believe these tests are running on Firefox 57 based on what I see from the logs. Firefox had a known bug, fixed in 59, where double clicking through selenium does not work. Tile view is going to change to pin participants using double click. What is your advice? Skip the test on Firefox?

damencho:
Skip them for now. We need to test latest FF whether it works, there were two issues we were waiting to be fixed to update from 57, one was fixed in 61 and one webrtc crash was recently fixed or planned to be fixed in latest and ESR but I need to further check and test selenium ...

virtuacoplenny:
@damencho, can you confirm for me that FF has been updated? If so I'll remove the change here to ignore FF.

damencho:
Yes, Firefox had been updated on the grid:
$ firefox --version
Mozilla Firefox 64.0
And the gecko driver is: 0.23.0

PR #220, opened 2018-12-17
fix(etherpad): verify etherpad can be edited before editing

virtuacoplenny:
There can be a delay between the etherpad inner body being
displayed and being editable.


damencho:
Unrelated failures.

PR #222, opened 2018-12-20
ref(shared-video): use a wrapper on click to wait for clickability

virtuacoplenny:
Often times tests will fail because a menu button is attempted
to be clicked before it is visible or clickable. Create a method
that encapsulates clicking with checks for clickability.


virtuacoplenny:
Not sure if this is a desired approach, and not sure if it's desired what the timeouts should be. I also think this method probably makes sense on the participant object itself, but left it in TestUtils as adding functionality to __Utils seems to be the norm.

PR #226, opened 2018-12-27
fix(lock-room): wait for password submit to be clickable

virtuacoplenny:
enterParticipantInLockedRoom is being flaky right now.
Screenshots show the password input is not cleared after
attempting to submit a known bad password, indicating
maybe the known password was never submitted.


virtuacoplenny:
If this fixes the flakiness, I'll probably just change all clicks to use the new util. I've been waiting on the page object PR to go in first but I'd rather not wait anymore if this does resolve flakiness.

virtuacoplenny:
Renamed the branch so it can run with the jitsi-meet PR that triggers the flakiness consistently. New PR is #227.

PR #227, opened 2018-12-27
fix(lock-room): wait for password submit to be clickable

virtuacoplenny:
enterParticipantInLockedRoom is being flaky right now.
Screenshots show the password input is not cleared after
attempting to submit a known bad password, indicating
maybe the known password was never submitted.


virtuacoplenny:
Jenkins test this please.

PR #229, opened 2019-01-10
Dial in test

damencho:
Depends on https://github.com/jitsi/jitsi-meet/pull/3778

### DIFF
--- a/src/test/java/org/jitsi/meet/test/DialInAudioTest.java
+++ b/src/test/java/org/jitsi/meet/test/DialInAudioTest.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright @ 2018 8x8 Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.meet.test;
+
+import com.google.gson.*;
+import org.apache.http.*;
+import org.apache.http.client.config.*;
+import org.apache.http.client.methods.*;
+import org.apache.http.impl.client.*;
+import org.apache.http.util.*;
+import org.jitsi.meet.test.pageobjects.web.*;
+import org.jitsi.meet.test.util.*;
+import org.jitsi.meet.test.web.*;
+import org.testng.*;
+import org.testng.annotations.*;
+
+import java.net.*;
+import java.text.*;
+
+import static org.testng.Assert.*;
+
+/**
+ * A simple tests that enters a conference, reads the dial-in info and executes
+ * a REST http call, which is suppose to add a dial-in participant in the call.
+ * We wait for some time for new participant, then wait for its audio and we
+ * kick it and consider this test successful.
+ *
+ * @author Damian Minkov
+ */
+public class DialInAudioTest
+    extends WebTestBase
+{
+    /**
+     * Name of the system property which holds the url to execute which will
+     * create a dial-in participant.
+     * The url must be a valid url, with properly escaped parameters.
+     * The url can contain {0}, to pass the conference pin.
+     * For example: {"pin":"{0}"}, needs to be: %7B%22pin%22%3A%22{0}%22%7D
+     */
+    private static final String DIAL_IN_PARTICIPANT_REST_URL
+        = "dialIn.rest.url";
+
+    private String dialInPin = null;
+
+    private String restURLString = null;
+
+    @Override
+    public boolean skipTestByDefault()
+    {
+        return true;
+    }
+
+    /**
+     * Creates a web participant to enter a room and read the dial-in info.
+     */
+    @Test
+    public void enterAndReadDialInPin()
+    {
+        this.restURLString = System.getProperty(DIAL_IN_PARTICIPANT_REST_URL);
+        Assert.assertTrue(
+            this.restURLString != null,
+            "REST Url missing. Pass it using " +
+                "-D" + DIAL_IN_PARTICIPANT_REST_URL + "=");
+
+        // join in a room
+        ensureOneParticipant();
+
+        WebParticipant participant = getParticipant1();
+
+        if (!MeetUtils.isDialInEnabled(participant.getDriver()))
+        {
+            throw new SkipException(
+                "No dial in configuration detected. Disabling test.");
+        }
+        InfoDialog infoDialog = participant.getInfoDialog();
+        infoDialog.open();
+
+        // get the dial-in pin
+        String displayedPin = infoDialog.getPinNumber();
+        assertTrue(displayedPin.length() > 1);
+
+        infoDialog.close();
+
+        // we need to remove the last character from the pin which is '#'
+        dialInPin = displayedPin.substring(0, displayedPin.length() - 1);
+    }
+
+    /**
+     * Executes the rest configured web call, replacing the dial-in pin the url
+     * and waits 10 seconds for the participant to join.
+     */
+    @Test(dependsOnMethods = { "enterAndReadDialInPin" })
+    public void enterDialInParticipant()
+    {
+        try
+        {
+            URI restURI = new URI(
+                MessageFormat.format(
+                    this.restURLString,
+                    this.dialInPin));
+
+            RequestConfig defaultRequestConfig = RequestConfig.custom()
+                .setSocketTimeout(5000)
+                .setConnectTimeout(5000)
+                .setConnectionRequestTimeout(5000)
+                .build();
+            CloseableHttpClient httpclient = HttpClientBuilder.create()
+                .setDefaultRequestConfig(defaultRequestConfig).build();
+
+            HttpHost targetHost = new HttpHost(
+                restURI.getHost(),
+                restURI.getPort(),
+                restURI.getScheme());
+
+            HttpGet httpget = new HttpGet(restURI);
+            try (CloseableHttpResponse response = httpclient.execute(
+                targetHost, httpget))
+            {
+                if (response.getStatusLine().getStatusCode() != 200)
+                {
+                    fail("REST returned error:" + response.getStatusLine());
+                }
+
+                HttpEntity entity = response.getEntity();
+                String value = EntityUtils.toString(entity);
+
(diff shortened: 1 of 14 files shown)


src/test/java/org/jitsi/meet/test/DialInAudioTest.java
```diff
@@ -0,0 +1,172 @@
...
+        if (!MeetUtils.isDialInEnabled(participant.getDriver()))
+        {
+            throw new SkipException(
+                "No dial in configuration detected. Disabling test.");
+        }
+        InfoDialog infoDialog = participant.getInfoDialog();
+        infoDialog.open();
+
+        // get the dial-in pin
+        String displayedPin = infoDialog.getPinNumber();
```
virtuacoplenny: Should this method remove the "#" from the pin?

src/test/java/org/jitsi/meet/test/DialInAudioTest.java
```diff
@@ -0,0 +1,172 @@
...
+                .build();
+            CloseableHttpClient httpclient = HttpClientBuilder.create()
+                .setDefaultRequestConfig(defaultRequestConfig).build();
+
+            HttpHost targetHost = new HttpHost(
+                restURI.getHost(),
+                restURI.getPort(),
+                restURI.getScheme());
+
+            HttpGet httpget = new HttpGet(restURI);
```
virtuacoplenny: For the sake of page objects, should this request logic be hidden behind a method on the participant or even in the infodialog? I was thinking the ideal is for the tests to read like english as much as possible and for test logic to be hidden away somewhere else.

src/test/java/org/jitsi/meet/test/DialInAudioTest.java
```diff
@@ -0,0 +1,172 @@
...
+        assertTrue(displayedPin.length() > 1);
+
+        infoDialog.close();
+
+        // we need to remove the last character from the pin which is '#'
+        dialInPin = displayedPin.substring(0, displayedPin.length() - 1);
+    }
+
+    /**
+     * Executes the rest configured web call, replacing the dial-in pin the url
```
virtuacoplenny: I think there might be typos in this comment.

src/test/java/org/jitsi/meet/test/pageobjects/web/WebRemoteParticipant.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright @ 2018 8x8 Pty Ltd
```
virtuacoplenny: 2019?

src/test/java/org/jitsi/meet/test/base/Participant.java
```diff
@@ -543,4 +543,23 @@ public void saveHtmlSource(File outputDir, String fileName)
      * @return a representation of the filmstrip of this participant.
      */
     public abstract Filmstrip<? extends Participant> getFilmstrip();
+
+    /**
+     * @return a list of representation of the remote participants thumbnails.
+     */
+    public abstract List<RemoteParticipant<T>> getRemoteParticipants();
```
virtuacoplenny: Should this method be renamed to include something about it being a thumbnail representation?

src/test/java/org/jitsi/meet/test/pageobjects/web/WebRemoteParticipant.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Opens the remote menu and waits for the desired link to appear.
+     * @param linkClassname the class name to use identifying the link.
+     */
+    private void clickOnRemoteMenuLink(String linkClassname)
+    {
+        // Open the remote video menu
+        WebElement cntElem
```
virtuacoplenny: What is cnt? I'd prefer this be named to what it is, without abbreviation, because I don't think cnt is a common abbreviation.

src/test/java/org/jitsi/meet/test/pageobjects/web/WebRemoteParticipant.java
```diff
@@ -0,0 +1,96 @@
...
+            = driver.findElement(By.id("participant_" + this.getEndpointId()));
+        String remoteVideoMenuButtonXPath
+            = TestUtils.getXPathStringForClassName(
```
virtuacoplenny: Usage of this helper is off to me, considering this might be the only place it gets used, meaning it has inconsistent usage so I think we should use it more or kill it. You can probably kill this helper by changing what By you're using. Without changing jitsi-meet more, you can just look for By.classname (or maybe use an xpath but check it includes the class). Ideally you try to keep selectors as concise as possible so that when the dom/layout changes the tests don't have to be modified, and tying to a span adds one more reason to change.

src/test/java/org/jitsi/meet/test/pageobjects/web/WebRemoteParticipant.java
```diff
@@ -0,0 +1,96 @@
...
+            driver,
+            "//ul[@class='popupmenu']//a[contains(@class, '"
+                + linkClassname + "')]",
+            5);
+
+        // click the button
+        MeetUIUtils.clickOnElement(driver,
+            "ul.popupmenu a." + linkClassname, true);
+
+        // wait for confirm muting to display
```
virtuacoplenny: Probably need to update these comments because they're currently mute specific.

src/test/java/org/jitsi/meet/test/pageobjects/web/WebRemoteParticipant.java
```diff
@@ -0,0 +1,96 @@
...
+        TestUtils.waitForDisplayedElementByXPath(
+            driver,
+            "//button[@id='modal-dialog-ok-button']",
+            5);
+
+        // confirm muting
+        MeetUIUtils.clickOnButton(driver,
+            "modal-dialog-ok-button", true);
```
virtuacoplenny: Make a generic modal page object? Out of scope to do so?

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
```diff
@@ -65,6 +67,12 @@
     public static final String ICE_CONNECTED_CHECK_SCRIPT =
         "return APP.conference.getConnectionState() === 'connected';";
 
+    /**
+     *  The javascript code which returns an array of remote participant IDs.
+     */
+    public static final String GET_REMOTE_PARTICIPANT_IDS =
+        "return APP.conference._room.getParticipants().map(p => p._id);";
```
virtuacoplenny: Do you think this is the best approach for the given desired changes? I'm generally not a fan of executing JS like this when it is avoidable, that way the end-to-end act more like a regular user. I think you can get the same info by scraping the dom, although then you rely on a stateful dom...so using JS seems okay too considering there's a lot of precedent for it already.

src/test/java/org/jitsi/meet/test/DialInAudioTest.java
damencho: Make sense, I will change it that way, it will not break other tests.

src/test/java/org/jitsi/meet/test/DialInAudioTest.java
damencho: Well, this is very specific for the current test. Initially had even named the test with REST in the name...

src/test/java/org/jitsi/meet/test/base/Participant.java
damencho: Any ideas ... do we have something similar on the other page objects?

src/test/java/org/jitsi/meet/test/pageobjects/web/WebRemoteParticipant.java
damencho: Haha, no idea :) I just moved the code. I will change it with something more readable.

src/test/java/org/jitsi/meet/test/pageobjects/web/WebRemoteParticipant.java
damencho: I will look at it, tomorrow to killing it.

src/test/java/org/jitsi/meet/test/pageobjects/web/WebRemoteParticipant.java
damencho: Ups.

paweldomas:
@damencho just replying to request for review from me: I think it's in good hands and it will be enough if Lenny looks at it, so don't wait for my review and just merge when he approves it.

src/test/java/org/jitsi/meet/test/pageobjects/web/WebRemoteParticipant.java
damencho: Will think of it ... I see we use for example modal-dialog-ok-button in 3 places at the moment...

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
damencho: Well, we are getting our own id using JS APP.conference ... so using the same way and getting remotes, I don't think is a problem.

src/test/java/org/jitsi/meet/test/base/Participant.java
virtuacoplenny: There really isn't a good existing precedent for interactions with remote participant thumbnails; from what I know MeetUIUtils is still used, which takes in an endpoint id. Existing page objects for the most part deal with UI outside of the filmstrip and not dynamically generated like the remote thumbnails.

I think it comes down to what you want the RemoteParticipant to represent. Do you want it to be more of an encompassing class that includes remote participant data, and maybe other helper methods related to the data, as well as DOM interactions? I'm feeling like right now you want it to instead represent the thumbnail itself, considering the javadoc.

virtuacoplenny: Then change the javadoc not to say it's returning a representation of the thumbnail?

damencho:
@virtuacoplenny Can you make another round of review. The changes for the modal dialog are in a separate commit and the rest of the changes addressing your comments are in the other commit. Thanks.

src/test/java/org/jitsi/meet/test/pageobjects/web/ModalDialogHelper.java
```diff
@@ -0,0 +1,63 @@
...
+     * Waits 5 seconds for the ok button to appear in the modal dialog.
+     * @param driver the page driver.
+     */
+    public static void waitForDisplayedOKButton(WebDriver driver)
+    {
+        TestUtils.waitForDisplayedElementByXPath(
+            driver,
+            "//button[@id='" + OK_BUTTON +"']",
+            5);
+    }
```
virtuacoplenny: TestUtils#click does this already.

damencho: I just updated it.

PR #230, opened 2019-01-31
fix(sharedvideo): use js to click on modal

virtuacoplenny:


zbettenbuk:
Merging it as it's needed for a test to come back green and the test failures are clearly not related to this change

PR #236, opened 2019-03-28
Do not fail dial-in test if there is a browser problem.

damencho:


src/test/java/org/jitsi/meet/test/DialInAudioTest.java
```diff
@@ -57,6 +57,15 @@
...
     private String restURLString = null;
 
+    /**
+     * We change this once the user had joined. We want to avoid situations
+     * where the grid has no available nodes, or the browser crashes/fails to
+     * load and to report a dial in failure in this case.
+     * When the value is false after we called <tt>ensureOneParticipant</tt>
+     * we skip all checks and return success.
+     */
+    private boolean userJoined = false;
```
paweldomas: Is it the case when there's the error logged about "page load failed" or something like that ?

damencho: Yes, exactly.

paweldomas: Then I'm not very happy about fixing just one test while in fact every single test can be affected by this. The worst is that it actually hides the failure.

paweldomas: I've been wondering if it's possible to rerun the test in such case ? That would be ideal. We wouldnt have to rerun the PR tests

damencho: I can try adding that to retry every participant entering if it fails to join ... just one retry, wdyt?

damencho: Well this test is a special case as we are running this as a monitoring and it reports failure in another component when the problem is just chrome not loading.
And this test is not ran by default and it needs special setup to run (a rest api to add a participant).

paweldomas: Oh I see.

I think it would be awesome if there would be such retry. It will make PR tests much more stable

PR #250, opened 2019-07-05
feat: adds a test that checks the jvb performance.

gpolitis:


damencho:
Hey @gpolitis can we actually configure this for the PR tests, so we can see the PR tests succeed?
The config will not be a problem for the rest of the tests as that config will not be used.

gpolitis:
@damencho would it be okay to postpone this until the new bridge is fully merged, or, even better, to activate this new test for the new jvb merge PR? the old bridge lacks some changes and and it currently fails this new test

damencho:
yep, hum I wonder about activating it for new bridge PRs, @bgrozev has done some work there, he can chime in and say can we do that.

src/test/java/org/jitsi/meet/test/web/WebTestBase.java
```diff
@@ -372,6 +408,8 @@ private void ensureInfoDialogClosed()
            webParticipant.getInfoDialog().close());
     }
 
+    private Set<String> endpointIds = new HashSet<>();
```
bgrozev: This should be at the top above the constructor, right?

bgrozev: And a description of what it holds would be nice

bgrozev:
I'm not sure I understand this correctly. The idea is to go over the jvb logs and fail specific tests based on that? For example, the "lock room" test might fail because of something in the jvb logs? Wouldn't it make more sense to collect all the logs after the torture run and analyze them separately?

bgrozev:
For the tests on the new bridge I added the target branch as a parameter to the scripts called by the `jitsi-videobridge-PR-test` job. That script checks if the target is 'nlj-jvb'. I don't remember exactly what it does differently, but it should be possible to make it run a specific branch of jitsi-meet-torture too.

gpolitis:
> I'm not sure I understand this correctly. The idea is to go over the jvb logs and fail specific tests based on that?

yeah, right now it only works with the time series but eventually we may wish to go through the generic jvb logs as well (and search, for example, for dropped packets)

> For example, the "lock room" test might fail because of something in the jvb logs? Wouldn't it make more sense to collect all the logs after the torture run and analyze them separately?

damencho also suggested this in the beginning and we can definitely do that; there are a couple of reasons why i implemented this as a per test final/last step; 1st because the failure may be related to the specific test that is run and 2nd in the future, if something is hard to repro, we may even decide to take a heap dump to help us identify a situation 

why do you guys think a check in the end makes more sense? we can easily change the code if there are advantages to that approach

PR #254, opened 2019-08-18
Give time for the kick to propagate in dial-in tests.

damencho:


src/test/java/org/jitsi/meet/test/DialInAudioTest.java
```diff
@@ -214,6 +214,10 @@ public void cleanupClass()
             if (participant != null)
             {
                 participant.getRemoteParticipants().get(0).kick();
+
+                // give time to click kick dialog confirm
+                // and send the kick command
+                waitMillis(1000);
```
paweldomas: is there no better way? not possible to wait for something?

damencho:
@paweldomas what about now? This looks better.

src/test/java/org/jitsi/meet/test/web/WebParticipant.java
```diff
@@ -401,7 +401,7 @@ public void waitForParticipants(int n)
         waitForCondition(
             () -> (Boolean) executeScript(
                 "return APP.conference"
-                    + ".listMembers().length >= " + n + ";"),
+                    + ".listMembers().length === " + n + ";"),
```
paweldomas: I'm wondering if this could break other tests which may depend on that, but we'll see...

damencho: It is the same test, waiting for 1 participant, no other tests use this :)

damencho: Thanks.

PR #257, opened 2019-10-08
Malleus tweaks

gpolitis:


src/test/java/org/jitsi/meet/test/web/WebTestBase.java
```diff
@@ -383,7 +383,7 @@ private void ensureInfoDialogClosed()
      * @param options the options to be used when creating the participant.
      * @return the participant which was created
      */
-    private WebParticipant joinParticipantAndWait(
+    public WebParticipant joinParticipantAndWait(
```
paweldomas: I'm not a big fan of this change. What's the problem with using `joinConference` ?

paweldomas: I may be wrong but it seems that 'ensureOneParticipant' method will do the same thing without exposing private method

paweldomas: Oh you need an index there. The at least maybe make the method protected instead of public.

gpolitis: Thanks @paweldomas. I talked to @damencho and he's not happy with this change either. He suggested to introduce a joinNextParticipant method and I'll try to implement that.

src/test/java/org/jitsi/meet/test/web/WebTestBase.java
```diff
@@ -278,35 +278,43 @@ private WebParticipant joinParticipant(
...
-
-            // Adds a print in the console/selenium-node logs
-            // useful when checking crashes or failures in node logs
-            p.executeScript(
-                    "console.log('--- Will start test:"
-                        + getClass().getSimpleName() + "')");
+        }
+        else
+        {
+            p = participants.createParticipant("", options);
```
paweldomas: what happens here?

paweldomas: why call create participant again if it exists?

gpolitis: @paweldomas this else clause belongs to the outer if statement that checks the value of the index. This branch executes when `index < 0` which is what `joinNextParticipant` passes down to this method. We use this newly introduced method in malleus where we don't care about what index a participant is placed at, but we need to create arbitrarily many participants.

paweldomas: Just leaving this as a note here that we have discussed with George that he will make an attempt to auto select the next index instead of allowing the case with empty prefix.

PR #258, opened 2019-12-30
Adjust fixnodes and mutatenodes in case JITSI_SSH_CONFIG isn't $HOME/ssh/config.d/jitsi64.config.

JonathanLennox:


scripts/fixnodes.sh
```diff
@@ -7,7 +7,12 @@ then
   exit 1
 fi
 
-SSH="ssh -F $HOME/.ssh/config.d/jitsi64.config"
+if [ -z "$JITSI_SSH_CONFIG" ]
+then
+  JITSI_SSH_CONFIG="$HOME/.ssh/config.d/jitsi64.config"
+fi
+
+SSH="ssh -F $JITSI_SSH_CONFIG"
```
gpolitis: just in case someone has a default config that works with our infra.

```suggestion
if [ -n "$JITSI_SSH_CONFIG" ]
then
  SSH="ssh -F $JITSI_SSH_CONFIG"
else
  SSH="ssh"
fi
```

scripts/mutatenodes.sh
```diff
@@ -29,14 +29,19 @@ then
   usage
 fi
 
+if [ -z "$JITSI_SSH_CONFIG" ]
+then
+  JITSI_SSH_CONFIG="$HOME/.ssh/config.d/jitsi64.config"
+fi
```
gpolitis: same as above?

scripts/fixnodes.sh
JonathanLennox: Do you in fact already have JITSI_SSH_CONFIG set to `$HOME/.ssh/config.d/jitsi64.config` in your environment? I don't want to break you.

gpolitis:
Superseded by https://github.com/jitsi/jitsi-meet-torture/pull/329.

PR #261, opened 2020-02-04
fix(ifame-api-test): disable analytics.

hristoterezov:


src/test/java/org/jitsi/meet/test/IFrameAPITest.java
```diff
@@ -52,6 +52,7 @@
         + "\"alwaysVisibleToolbar\":true,"
         + "\"p2p.enabled\":false,"
         + "\"disable1On1Mode\":true"
```
damencho: Don't you miss a ',' here.

hristoterezov: Fixed!

PR #268, opened 2020-03-31
Fix for Peer2Peer check

TheNetStriker:
Changed APP.conference.getConnectionState() to APP.conference.getP2PConnectionState() to fix Peer2Peer check.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/util/MeetUtils.java
+++ b/src/test/java/org/jitsi/meet/test/util/MeetUtils.java
@@ -36,7 +36,7 @@ public class MeetUtils
      * is in state 'connected'.
      */
     public static final String ICE_CONNECTED_CHECK_SCRIPT =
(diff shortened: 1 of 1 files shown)


TheNetStriker:
I've just signed the document with my e-mail: david@masshardt.ch

src/test/java/org/jitsi/meet/test/util/MeetUtils.java
```diff
@@ -36,7 +36,7 @@
      * is in state 'connected'.
      */
     public static final String ICE_CONNECTED_CHECK_SCRIPT =
-        "return APP.conference.getConnectionState() === 'connected';";
+        "return APP.conference.getP2PConnectionState() === 'connected';";
```
paweldomas, line 39: no, this script is supposed  to check if conference's ICE state is connected and not just  P2P

PR #277, opened 2020-05-20
Add security dialog tests

muscat1:


damencho:
@saghul @muscat1 Why did we remove the Info dialog and closing it? That breaks a lot of tests by taking the focus and not able to click on the screen?

damencho:
Like this:
image

saghul:
Clarified with @damencho. Tests do pass on master, after the security dialog was merged.

PR #281, opened 2020-05-26
fix #280

### LINKED ISSUE #280: test fails because language is not english.
It seems that the test is based on english, the element selecter is based on the language-spec selecter, so lot's of test fails because of unable to find the element.
for example `$("*[aria-label='Toggle mute audio']")` will find nothing, but `$("*[aria-label='静音 / 取消静音']")` will find the correct element on my computer.

vipcxj:
fix #280 
Force chrome and firefox web driver use English instead of system language.
The code for chrome has test on my computer. The chrome version is 83.0.4103.61. The os system is win 10.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
+++ b/src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
@@ -208,7 +208,7 @@ public class WebParticipantFactory
             final ChromeOptions ops = new ChromeOptions();
             // Force chrome to use English instead of system language.
             // At least work at version 83.0.4103.61 on win 10
-            ops.addArguments("--lang=en");
+            ops.addArguments("lang=en");
             ops.addArguments("allow-insecure-localhost");
             ops.addArguments("use-fake-ui-for-media-stream");
             ops.addArguments("use-fake-device-for-media-stream");
--- a/src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
+++ b/src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
@@ -133,6 +133,9 @@ public class WebParticipantFactory
             }
 
             FirefoxProfile profile = new FirefoxProfile();
+            // Force firefox to use English instead of system language.
+            // Not test because of having not firefox installed.
+            profile.setPreference("intl.accept_languages","en");
             profile.setPreference("media.navigator.permission.disabled", true);
             // Enables tcp in firefox, disabled by default in 44
             profile.setPreference("media.peerconnection.ice.tcp", true);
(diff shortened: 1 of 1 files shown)


src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
```diff
@@ -203,6 +206,9 @@ else if (participantType == ParticipantType.edge)
             logPrefs.enable(LogType.BROWSER, Level.ALL);
 
             final ChromeOptions ops = new ChromeOptions();
+            // Force chrome to use English instead of system language.
+            // At least work at version 83.0.4103.61 on win 10
+            ops.addArguments("--lang=en");
```
damencho, line 211: I think these arguments need to be without `--`

vipcxj: It should with --

vipcxj: There is another solution.
```java
            Map<String, String> lops = new HashMap<>();
            lops.put("intl.accept_languages", "en,en_US");
            ops.setExperimentalOption("prefs", lops);
```
It work on my computer, too.
But `addArguments` is short, so I choose that.

damencho: Look at the rest of the arguments like `ops.addArguments("allow-insecure-localhost");`
On the command line that is `--allow-insecure-localhost`.

vipcxj: with `--` it just work. without `--` I can't promise it will still work.

damencho: You had tested it with torture and -- ? Or you just had tested it -- command line?

damencho: Or you are right, it does not matter https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/chrome/ChromeOptions.html#addArguments-java.util.List-

damencho: As you can see it the docs the examples are without --

vipcxj: I have test without `--` on windows, it also works, so I removed the `--` to keep the format.

vipcxj:
I found some other bugs when using this test framework on windows. Such as send text to the input sometimes crash because the input is invalid. I think it perhaps because the react update the component when the input is focused.

PR #282, opened 2020-05-27
support port which not default to 443

vipcxj:
IFrameAPITest failed because the wrong port. This pull request make test work on the host with a port other than 443, such as 8080.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/IFrameAPITest.java
+++ b/src/test/java/org/jitsi/meet/test/IFrameAPITest.java
@@ -63,7 +63,7 @@ public class IFrameAPITest
         String domain;
         try
         {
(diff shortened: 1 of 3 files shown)


vipcxj:
`FilmstripOnlyTest#testDisplaysDeviceSelection` also failed, too. I dig into it and found `APP_ID` is undefined. and window.location.search and window.location.params are all empty. but a url with config hash(https://localhost:8080/XXX#interfaceConfig.filmStripOnly=true) works well, `FilmstripOnlyTest#testLoadsOnlyTheFilmstrip` success and the OnlyTheFilmstrip page shows.

Neustradamus:
@damencho: Can you look this PR?

PR #290, opened 2020-07-13
Reduces run time and adds check style

damencho:


damencho:
The failure is strange, will try do debug what is going on: `FAIL: Average processing time too high for node RtcpTermination: 169901.46818742293ns`

But in general, this reduces the time of the PR tests from 14-15 minutes to 10 min.

PR #291, opened 2020-07-14
Fix lock room test

damencho:


damencho:
We are trying a different approach https://github.com/jitsi/jitsi-meet-torture/pull/292, closing this for now.

PR #294, opened 2020-07-15
feat: Add grant moderator test

gimre:


saghul:
Jenkins please test this please.

saghul:
Jenkins please test this please.

gimre:
Jenkins please test this please.

damencho:
Hey @gimre is it possible to detect whether this functionality is available on the server? 
The idea is that if you run tests against deployment that is still an older version to skip the test, similar to what we do for Etherpad tests.

saghul:
I guess he could check for `APP.conference._room.grantXXX` and check if it's not undefined?

src/test/java/org/jitsi/meet/test/pageobjects/mobile/stats/TestConnectionInfo.java
```diff
@@ -47,6 +47,16 @@
     private static final String ID_CONFERENCE_JOINED_STATE
         = "org.jitsi.meet.conference.joinedState";
 
+    private static final String ID_GRANT_MODERATOR_ENABLED
```
saghul, line 50: Let's call this ID_GRANT_MODERATOR_AVAILABLE, since it's not something you can turn off.

gimre: Updated it, also in the JM PR!

PR #297, opened 2020-07-22
Add test for digits only room lock

saghul:
Depends on https://github.com/jitsi/jitsi-meet/pull/7372

### DIFF
--- a/src/test/java/org/jitsi/meet/test/LockRoomDigitsOnlyTest.java
+++ b/src/test/java/org/jitsi/meet/test/LockRoomDigitsOnlyTest.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright @ 2015 Atlassian Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.meet.test;
+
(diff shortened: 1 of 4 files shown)


src/test/java/org/jitsi/meet/test/LockRoomDigitsOnlyTest.java
```diff
@@ -0,0 +1,81 @@
...
+     */
+    @Test
+    public void lockRoomWithDigitsOnly()
+    {
+        ensureOneParticipant(
+            getJitsiMeetUrl().appendConfig(
+                "config.roomPasswordNumberOfDigits=5"));
+
+        // just in case wait
+        TestUtils.waitMillis(1000);
```
damencho, line 46: Is this needed? 
... You can do getParticipant1().waitToJoinMUC();

src/test/java/org/jitsi/meet/test/LockRoomDigitsOnlyTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+        // Set a valid numeric password.
+        securityDialog.addPassword("12345");
+        TestUtils.waitMillis(1000);
+        securityDialog.close();
+    
+        TestUtils.waitMillis(1000);
+        securityDialog.open();
+
+        assertTrue(securityDialog.isLocked());
```
damencho, line 79: I would love to get rid of all the waits ... We are trying to clear those, as during the years we have a lot of those (I personally am responsible for a lot of those) some without explanation ... and it is hard to understand after some time ...

damencho:
Other than the waits LGTM

damencho:
And the failure 😛

src/test/java/org/jitsi/meet/test/LockRoomDigitsOnlyTest.java
saghul: Sounds good, will do!

saghul:
> And the failure 😛

Hehe, the problem is that the tests will fail unless the config override works. So, do you have a suggestion on how to handle that?

damencho:
I'm merging with a plan to remove those waits as part of my next PR which adds some modifications to security dialog.

PR #303, opened 2020-08-14
feat: Adds one more lobby test case.

damencho:
When there are 3 participants in a call and lobby is enabled we want to make sure the call continues to have 3 participants.


damencho:
The first run of this test should fail, as it tests a bug which is not fixed.

PR #319, opened 2020-11-18
fix: Temporary wait for the avatar to download.

damencho:


damencho:
@saghul If you will be checking what is going on, we should drop this commit if you think you have a fix for quicker update the avatar downloading.

bgrozev (approved):

PR #321, opened 2020-11-25
fix: Add SLF4J logger implementation dependency

slykar:
When I was trying to run `./scripts/malleus.sh` with Chrome driver I got the following error:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

Adding the dependency fixes the issue and logs get printed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,7 +87,11 @@
             <version>3.0</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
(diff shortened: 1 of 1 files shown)


damencho:
Jenkins test this please.

damencho (approved):

PR #326, opened 2020-12-10
feat: Allow malleus to test instanceswith self-signed certificates.

gpolitis:


damencho (approved):

scripts/malleus.sh
```diff
@@ -92,6 +97,7 @@ mvn \
 -Dorg.jitsi.malleus.room_name_prefix=$ROOM_NAME_PREFIX \
 -Dorg.jitsi.malleus.regions=$REGIONS \
 -Dremote.address=$HUB_URL \
+-DallowInsecureCerts=$ALLOW_INSECURE_CERTS \
```
damencho, line 100: But wait

damencho: I was testing with web.participant1.allowInsecureCerts=true ... web.participant2.allowInsecureCerts=true

damencho (changes requested):

scripts/malleus.sh
damencho: Have you tested it, maybe we need a global/default one ...

damencho (approved):

PR #329, opened 2020-12-18
Run multiple Malleus endpoints on a single node, 

JonathanLennox:
By mutating the Selenium grid appropriately.

### DIFF
--- a/scripts/malleus.sh
+++ b/scripts/malleus.sh
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+usage() {
+    echo "Usage: $0 [--conferences=CONFERENCES] [--participants=PARTICIPANTS] [--senders=SENDERS] [--audio-senders=AUDIO_SENDERS] [--senders-per-node=SENDERS_PER_NODE] [--receivers-per-node=RECEIVERS_PER_NODE] [--duration=DURATION] [--room-name-prefix=ROOM_NAME_PREFIX] [--hub-url=HUB_URL] [--instance-url=INSTANCE_URL] [--regions=REGIONS] [--skip-mutation]" >&2
+    exit 1
+}
+
 case $1 in
   --*)
     # new arg parsing code that includes default values for the different options.
@@ -12,14 +17,16 @@ case $1 in
         --participants) PARTICIPANTS=$optvalue;;
         --senders) SENDERS=$optvalue;;
         --audio-senders) AUDIO_SENDERS=$optvalue;;
+        --senders-per-node) SENDERS_PER_NODE=$optvalue;;
+        --receivers-per-node) RECEIVERS_PER_NODE=$optvalue;;
         --duration) DURATION=$optvalue;;
         --room-name-prefix) ROOM_NAME_PREFIX=$optvalue;;
         --hub-url) HUB_URL=$optvalue;;
         --instance-url) INSTANCE_URL=$optvalue;;
         --regions) REGIONS=$optvalue;;
+        --skip-mutation) SKIP_MUTATION=1;;
         *)
-          echo 'Usage: $0 [--conferences=CONFERENCES] [--participants=PARTICIPANTS] [--senders=SENDERS] [--audio-senders=AUDIO_SENDERS] [--duration=DURATION] [--room-name-prefix=ROOM_NAME_PREFIX] [--hub-url=HUB_URL] [--instance-url=INSTANCE_URL] [--regions=REGIONS]' >&2
-          exit 1
+          usage
           ;;
       esac
     done
@@ -42,6 +49,14 @@ case $1 in
 
     if [ -z "$AUDIO_SENDERS" ]; then
       AUDIO_SENDERS=$PARTICIPANTS
+    fi
+
+    if [ -z "$SENDERS_PER_NODE" ]; then
+      SENDERS_PER_NODE=1
+    fi
+
+    if [ -z "$RECEIVERS_PER_NODE" ]; then
+      RECEIVERS_PER_NODE=1
     fi
 
     if [ -z "$DURATION" ]; then
@@ -86,6 +101,12 @@ case $1 in
     INSTANCE_URL='https://meet.jit.si'
     ;;
 esac
+
+if [ -z "$SKIP_MUTATION" ]
+then
+    "$(dirname $0)"/mutatenodes.sh --hub-url="$HUB_URL" --num-senders="$SENDERS" \
+		   --send-node-max-sessions="$SENDERS_PER_NODE" --recv-node-max-sessions="$RECEIVERS_PER_NODE"
+fi
 
 mvn \
 -Dthreadcount=1 \
--- a/scripts/mutatenodes.sh
+++ b/scripts/mutatenodes.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+set -e
+
+SEND_APP_NAME='malleusSender'
+RECV_APP_NAME='malleusReceiver'
+
+usage() {
+    echo "Usage: $0 --hub-url=HUB_URL --num-senders=NUM_SENDERS --send-node-max-sessions=SEND_NODE_MAX_SESSIONS --recv-node-max-sessions=RECV_NODE_MAX_SESSIONS" >&2
+    exit 1
+}
+
+# new arg parsing code that includes default values for the different options.
+for arg in "$@"; do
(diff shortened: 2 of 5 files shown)


scripts/malleus.sh
```diff
@@ -12,14 +17,16 @@ case $1 in
...
         --senders) SENDERS=$optvalue;;
         --audio-senders) AUDIO_SENDERS=$optvalue;;
+        --senders-per-node) SENDERS_PER_NODE=$optvalue;;
+        --receivers-per-node) RECEIVERS_PER_NODE=$optvalue;;
         --duration) DURATION=$optvalue;;
         --room-name-prefix) ROOM_NAME_PREFIX=$optvalue;;
         --hub-url) HUB_URL=$optvalue;;
         --instance-url) INSTANCE_URL=$optvalue;;
         --regions) REGIONS=$optvalue;;
+        --skip-mutation) SKIP_MUTATION=1;;
```
bbaldino, line 27: What does `mutation` refer to?

bbaldino: Even after seeing the file below, I'm not sure it's clear

bbaldino: After re-reading it more I get it's about changing an existing mode, but I'm not sure what it implies to skip or not skip mutation.

JonathanLennox: It means to change the settings of the Selenium node, to configure it to have N browsers, set up to be senders or receivers.  Any suggestion for a better name?

bbaldino: Is this to say: selenium nodes have some default setting and this would cause the script to change the state of a node? Is the state permanent? Is it required if, say, `senders-per-node` is set to a number that doesn't match what the current node is set up for?

JonathanLennox: Yes, Selenium nodes (at least in our Grid setup) by default run only a single browser, and don't have the `applicationName` field set in their configuration.

The script changes the nodes to run either `senders-per-node` or `receivers-per-node` browsers, distinguishing them with `applicationName`.

The changes are permanent, currently, yes.  Do you think I should have the script revert them when it's done, instead?

bbaldino: Ok, I'm still confused by the semantics of `skip-mutation`, how does it work to skip mutation if the number of browsers on the node don't match what was requested?

JonathanLennox: Well, as long as there are sufficiently many nodes marked with the appropriate `applicationName` values, things will work.  That said, if they *don't* have `applicationName` values, it'll just fail.

My goal here was to let things still (kind of) work for people who might be using Selenium grids configured differently than ours (since this is in the open-source project), but perhaps this needs a little more thought?

bbaldino: Yeah I guess I'm missing context here--I'm not aware of all the semantics around `applicationName` either.  Are there applications we're talking about here other than selenium?

JonathanLennox: `applicationName` is just a (poorly-documented) way for a Selenium node to indicate an endpoint property more specific than the browser type.  When a client requests a browser, it can (optionally) indicate that it needs a browser that has the specified `applicationName` value configured for it.  (There's no semantics from Selenium itself, it's only a way of binding node configuration to client requests.)

This script is using it to distinguish "sender" nodes (i.e. ones that will launch at most `senders-per-node` browsers) from "receiver" nodes (that will launch at most `receivers-per-node` browsers).

bbaldino: Ok, I'll just consider myself here to ask the dumb questions then :) 

> This script is using it to distinguish "sender" nodes (i.e. ones that will launch at most senders-per-node browsers) from "receiver" nodes (that will launch at most receivers-per-node browsers).

What about endpoints that we want to send and receive?

JonathanLennox: "Sender" means send-and-receive; we don't have send-only nodes currently.  (Given that we're just exercising the jitsi-meet UI in this test, is there even a way to configure a node as send-only?)

bbaldino: So is it that: we get some configuration from the command line which has some amount of total senders and receivers, and then a setting of the maximum amount of senders per node and the maximum amount of receivers per node, and then the backend grabs a node and then:

1) Checks what it's current configuration is to see if it fits within senders-per-node and receivers-per-node, and then, if necessary, changes the amount of browsers spun up on that node?

or 
2) Just overrides whatever config exists to make this node conform to senders-per-node and receivers-per-node?

JonathanLennox: It's option 2. It doesn't do any checking.

I don't think "see if it fits within" is a good idea; one use case I was concerned about is when you discover that the limits you set for a previous run turned out to be too aggressive, and you need to set them lower.  (This is why I also have the script to monitor the Selenium load average in this patch set.)

scripts/mutatenodes.sh
```diff
@@ -0,0 +1,91 @@
...
+    else
+	SSH="ssh"
+    fi
+fi
+
+if [ -z "$SCP" ]
+then
+    if [ -r "$JITSI_SSH_CONFIG" ]
+    then
+	SCP="scp -F $JITSI_SSH_CONFIG"
```
gpolitis, line 52: I think there are tabs here? Maybe you can use 4 spaces instead (just for consistency with the rest of the file)

scripts/mutatenodes.sh
```diff
@@ -0,0 +1,91 @@
...
+if [ -z "$JITSI_SSH_CONFIG" ]
+then
+    JITSI_SSH_CONFIG="$HOME/.ssh/config.d/jitsi64.config"
+fi
+
+if [ -z "$SSH" ]
+then
+    if [ -r "$JITSI_SSH_CONFIG" ]
+    then
+	SSH="ssh -F $JITSI_SSH_CONFIG"
```
gpolitis, line 42: same as below

scripts/mutatenodes.sh
gpolitis: hmm it seems the file has a mixture of 4 spaces, 2 spaces and tabs? I don't have any preference, but maybe it's worth choosing one spacing style and apply it across all lines

JonathanLennox: Fixed.

gpolitis (dismissed):

scripts/monitornodes.sh
```diff
@@ -0,0 +1,42 @@
...
+
+if [ $# -gt 1 ]; then
+    HUB_URL=$1
+else
+    HUB_URL="http://localhost:4444/wd/hub"
+fi
+
+HUB_CONSOLE_URL="$(echo "$HUB_URL" | sed 's,/wd/hub,/grid/console,')"
+
+if [ -z "$JITSI_SSH_CONFIG" ]
```
gpolitis, line 16: The default path here is what I use in my dev env and I don't know if people are okay with such a default. We could remove it if we wanted to be less opinionated.

scripts/monitornodes.sh
```diff
@@ -0,0 +1,42 @@
...
+then
+    JITSI_SSH_CONFIG="$HOME/.ssh/config.d/jitsi64.config"
+fi
+
+if [ -z "$SSH" ]
+then
+    if [ -r "$JITSI_SSH_CONFIG" ]
```
gpolitis, line 23: How about simply SSH_CONFIG?

scripts/mutatenodes.sh
```diff
@@ -0,0 +1,108 @@
...
+            ;;
+    esac
+done
+
+if [ -z "$HUB_URL" -o -z "$NUM_SENDERS" -o -z "$SEND_NODE_MAX_SESSIONS" -o -z "$RECV_NODE_MAX_SESSIONS" ]
+then
+    usage
+fi
+
+if [ -z "$JITSI_SSH_CONFIG" ]
```
gpolitis, line 33: Just a reminder to update this script, If we decide to change anything in the monitor script

scripts/monitornodes.sh
JonathanLennox: This is specifically for connecting to the nodes you're using for Jitsi, which is why I named it that way (and also because that was the name I already had for the environment variable in my local config).  Presumably other ssh will use a different config...

scripts/monitornodes.sh
JonathanLennox: Any default, or no default, would be fine for me.

gpolitis (approved):

PR #330, opened 2021-01-07
Malleus with blips

gpolitis:
This PR extends the Malleus script with a `--max-disrupted-bribges-pct` command line switch that defines the percentage of bridges to _break_.

The specific implementation of the blip.sh script is left to the user, as it is tied to an infrastructure. We provide a placeholder script that the user needs to replace with a functional script. 

Depends on (and builds upon) https://github.com/jitsi/jitsi-meet-torture/pull/329


src/test/java/org/jitsi/meet/test/BlipTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.jitsi.meet.test.util.*;
+import org.jitsi.meet.test.web.*;
+
+/**
+ * A test that works exactly like the {@link MalleusJitsificus} but with blips
+ * enabled by default and some health checks that are periodically run on the
+ * endpoints. Also see {@link ConferenceMigrationTest}.
+ */
+public class BlipTest extends MalleusJitsificus
```
damencho, line 26: I don't understand what is actually tested? 
Are you expecting that BlipTest and ConferenceMigrationTest will run at the same time ... ?

damencho (approved):

damencho (approved):

PR #346, opened 2021-02-12
Add Join Delay (delay between participant joining) to Malleus.

JonathanLennox:


src/test/java/org/jitsi/meet/test/MalleusJitsificus.java
```diff
@@ -85,7 +87,9 @@
                 ? numParticipants
                 : Integer.parseInt(numAudioSendersStr);
 
-        int timeoutMs = 1000 * Integer.parseInt(System.getProperty(DURATION_PNAME));
+        int durationMs = 1000 * Integer.parseInt(System.getProperty(DURATION_PNAME));
+
+        int joinDelayMs = 1000 * Integer.parseInt(System.getProperty(JOIN_DELAY_PNAME));
```
bgrozev, line 92: The doc above says that it is already in milliseconds?

JonathanLennox: Whoops, you're right.

bgrozev (approved):

PR #349, opened 2021-02-18
Fixes malleus shell function calls

gpolitis:


JonathanLennox (approved):
This is probably a bash vs zsh difference, but so long as it still works with bash too, it's fine with me.

PR #350, opened 2021-02-18
Blips with load test fixes

gpolitis:
The actual fix is here 5c58e17f6c326eda2924a711c41a736226bf06ee and consists of bumping the getBridgeIp timeout to 30s. The rest is some minor refactoring and cleanup.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,7 +90,7 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-simple</artifactId>
-            <version> 1.7.30</version>
+            <version>1.7.30</version>
         </dependency>
     </dependencies>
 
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -90,7 +90,7 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-simple</artifactId>
-            <version> 1.7.30</version> 
+            <version> 1.7.30</version>
```
JonathanLennox, line 93: Should remove whitespace before the version number too, while you're cleaning this up.

gpolitis: Thank you Jonathan, fixed.

JonathanLennox (approved):

PR #354, opened 2021-02-26
Fail malleus if the participant fails

gpolitis:
I refactored the code so that we can pull any exceptions that happened from the participant thread, and fail the overall test when it finishes.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/MalleusJitsificus.java
+++ b/src/test/java/org/jitsi/meet/test/MalleusJitsificus.java
@@ -37,7 +37,6 @@ public class MalleusJitsificus
      */
     private static final String INPUT_VIDEO_FILE
         = "resources/FourPeople_1280x720_30.y4m";
-    private static final int DISABLE_FAILURE_DETECTION = -1;
 
     public static final String CONFERENCES_PNAME
         = "org.jitsi.malleus.conferences";
@@ -67,6 +66,13 @@ public class MalleusJitsificus
     private final Phaser allHungUp = new Phaser();
 
     private CountDownLatch bridgeSelectionCountDownLatch;
+
+    // The participant threads put the IP of their bridge in this set.
+    private final Set<String> bridgeSelection = ConcurrentHashMap.newKeySet();
+
+    // The test thread creates this set.
+    // The participant threads check if the IP of their bridge is in this set.
+    private Set<String> bridgesToFail;
 
     @DataProvider(name = "dp", parallel = true)
     public Object[][] createData(ITestContext context)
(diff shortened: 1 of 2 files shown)


src/test/java/org/jitsi/meet/test/MalleusJitsificus.java
```diff
@@ -170,69 +176,61 @@ public void testMain(
...
-                    {
-                        t.join();
-                    }
-                }
+                // TODO we should cancel the futures since test has failed
+                errors.add(new Exception("Failed to disrupt the bridges.", e));
             }
         }
 
-        int minFailureTolerance = Integer.MAX_VALUE;
```
gpolitis, line 217: The logic that checks if the participant failed "too much" was moved into the participant thread code and changed into throwing an exception (that is now handled).

src/test/java/org/jitsi/meet/test/MalleusJitsificus.java
```diff
@@ -170,69 +176,61 @@ public void testMain(
...
+            catch (ExecutionException e)
+            {
+                errors.add(e.getCause());
             }
         }
 
-        if (disruptBridges && minFailureTolerance < 0)
+        if (!errors.isEmpty())
         {
-            throw new Exception("Minimum failure tolerance is less than 0");
```
gpolitis, line 230: There are timeout exceptions that were not covered by this check before.

src/test/java/org/jitsi/meet/test/MalleusJitsificus.java
```diff
@@ -170,69 +176,61 @@ public void testMain(
...
+            throw new Exception("Failed with multiple errors. Throws the primary.", errors.get(0));
         }
     }
 
-    private class ParticipantThread
-        extends Thread
+    private class MalleusTask
```
gpolitis, line 232: Maybe this needs to be renamed?

src/test/java/org/jitsi/meet/test/MalleusJitsificus.java
```diff
@@ -241,15 +239,15 @@ public void testMain(
...
         WebParticipant participant;
-        public int failureTolerance;
         private String bridge;
 
-        public ParticipantThread(
+        public MalleusTask(
             int i, JitsiMeetUrl url, long durationMs, long joinDelayMs,
             boolean muteVideo, boolean muteAudio, String region,
-            int initialFailureTolerance)
+            boolean enableFailureDetection)
```
gpolitis, line 250: The boolean that Jonathan had introduced in an earlier version of this class was good enough.

damencho (approved):

PR #366, opened 2021-03-15
Update WebParticipant.java

ahmetkarakaya:
in jitsi config.js file, the default value of  pcStatsInterval is 10000, so current value 1500 should be changed into 10000.

--- config.js--- 
  // The interval at which PeerConnection.getStats() is called. Defaults to 10000, 
    // pcStatsInterval: 10000,


ahmetkarakaya:
> Hi, thanks for your contribution!
> If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(.

I have signed the document for ICLA

damencho:
This was changed on purpose, otheyrwose some tests fail and for some tests it takes more than the needed time to execute which leads to overall more time in waiting PR tests to finish.

PR #367, opened 2021-03-17
Add code to switch speakers in Malleus.

JonathanLennox:
Makes malleus execution asynchronous to make this cleaner.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/MalleusJitsificus.java
+++ b/src/test/java/org/jitsi/meet/test/MalleusJitsificus.java
@@ -176,16 +176,16 @@ public class MalleusJitsificus
         String[] regions, float blipMaxDisruptedPct)
         throws Exception
     {
-        Future<?>[] futures = new Future[numberOfParticipants];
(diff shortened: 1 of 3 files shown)


src/test/java/org/jitsi/meet/test/MalleusJitsificus.java
```diff
@@ -278,8 +318,38 @@ public void run()
...
+        public void start(ScheduledExecutorService pool)
+        {
+            this.pool = pool;
+
+            started = pool.schedule(this::join, joinDelayMs, TimeUnit.MILLISECONDS);
+            complete = pool.schedule(this::finish, joinDelayMs + durationMs, TimeUnit.MILLISECONDS);
+
+            if (enableFailureDetection)
+            {
+                long healthCheckIntervalMs = 5000; // Configure this?
```
gpolitis, line 336: The default value is good enough for our use-case and we wouldn't want to tweak it but I agree it would be nice to have this configurable just because arbitrary constants mixed in the code are rarely a good idea.

gpolitis (approved):

PR #381, opened 2021-04-19
feat(share-video) update share video tests to include generic video s…

quitrk:
…haring

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,8 @@ target/
 # PSNR test sequences and output files
 resources/psnr/output/
 resources/psnr/test_sequences/
+
+# VSCode files
+.classpath
+.project
+.settings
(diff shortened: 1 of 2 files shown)


damencho:
Is youtube sharing still supported?

damencho:
If so, let's keep both tests then.

quitrk:
> If so, let's keep both tests then.

they're both tested...first link is still a youtube link

damencho (approved):

damencho:
Yep ... I missewd that, mind that this needs to be merged at the same time as the jitsi-meet change. 

Why did you drop the `volumeTest `?

quitrk:
> Yep ... I missewd that, mind that this needs to be merged at the same time as the jitsi-meet change.
> 
> Why did you drop the `volumeTest `?

we're not syncing the video volume anymore

PR #383, opened 2021-05-04
fix: Skips waiting FF to become dominant speaker.

damencho:
FF uses a constant beep as an audio input which cannot trigger dominant speaker change.


bgrozev (approved):

damencho:
Jenkins test this please.

PR #389, opened 2021-06-14
jwt tests

damencho:


src/test/java/org/jitsi/meet/test/misc/SingleModeratorTest.java
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * @author Damian Minkov
+ */
+public class SingleModeratorTest
+    extends WebTestBase
+{
+    /**
+     * Required configuration keys.
+     */
+    private static final String TENANT_NAME_PNAME = "org.jitsi.misc.single.tenant_name";
```
gpolitis, line 40: Wouldn't it make sense to extract a base class for these new tests in order to avoid repetition?

src/test/java/org/jitsi/meet/test/misc/SingleModeratorTest.java
```diff
@@ -0,0 +1,107 @@
...
+    {
+        super.setupClass();
+
+        tenantName = System.getProperty(TENANT_NAME_PNAME);
+        roomName = System.getProperty(MODERATOR_ROOM_NAME_PNAME);
+        moderatorToken = System.getProperty(MODERATOR_TOKEN_PNAME);
+        nonModerator1Token = System.getProperty(NON_MODERATOR1_TOKEN_PNAME);
+        nonModerator2Token = System.getProperty(NON_MODERATOR2_TOKEN_PNAME);
+
+        if (Utils.checkForMissingConfig(tenantName, roomName, moderatorToken, nonModerator1Token, nonModerator2Token))
```
gpolitis, line 66: If we extract a base class this would have to be specialized.

gpolitis (approved):

PR #398, opened 2021-07-21
test(AudioVideoModerationTest)

Calinteodor:


src/test/java/org/jitsi/meet/test/AudioVideoModerationTest.java
```diff
@@ -0,0 +1,175 @@
+package org.jitsi.meet.test;
```
damencho, line 1: Add the license headers please.

damencho (changes requested):
One more case you can cover:
- moderation is enabled
- participant has requested to unmute (participant 3)
- the moderator reloads, in the single-moderator use case we are testing here, a random participant (maybe participant 2) will become moderator, make sure that new moderator can allow unmuting.

Calinteodor:
> One more case you can cover:
> 
> * moderation is enabled
> * participant has requested to unmute (participant 3)
> * the moderator reloads, in the single-moderator use case we are testing here, a random participant (maybe participant 2) will become moderator, make sure that new moderator can allow unmuting.

If moderator(participant 1) reloads this means participant 2 will become moderator and unmute will be available for all participants audio/video

damencho:
Jenkins, test this please.

damencho (changes requested):
So I have few things about the PR. 

1. Please remove all `TestUtils.waitMillis(1000);` we need to use those as a last resort. This adds a lot of wait and is a lot for PR testing :) You must wait X seconds till something happens or appears. For example in raiseHandToSpeak you click unmute, and wait 2000 seconds to check for the notification. You need to wait up to two seconds till the notification appears, there are plenty of examples in the code about that. Maybe like https://github.com/jitsi/jitsi-meet-torture/blob/527babf69b7b09a309a280684325ff727897b326/src/test/java/org/jitsi/meet/test/pageobjects/web/Notifications.java#L82
2. Please use separate methods for different tested cases, check the lobby test for example, how to order them.
3. - First test case, the happy one: 
    3 participants. Participant 1 enable av moderation. Participant 2 asks to unmute. Participant 1 sees the dialog, and allows it and checks that the notification disappeared (this is with participant pane closed). 
    Participant 3 asks to unmute, and moderator allowes it from within the participant pane (this way we test both UI components). In both cases make sure that the participant can unmute.
- Second test case: 
    You can hangup participant 2 and 3 and join them again and make sure they started muted. Ask to unmute from both, then hangup the moderator and join him again. The new moderator will be seeing only one notification to allow someone to unmute. The new moderator can be participant2 or participant3 not guranteed which one - your latest PR test failed because the other one was moderator, I think http://pr-reports.jitsi.org/jitsi-meet-PR-test/9929/ff-chrome-reports/html/.

So put case1 and case2 in different methods.

Calinteodor:
> So I have few things about the PR.
> 
> 1. Please remove all `TestUtils.waitMillis(1000);` we need to use those as a last resort. This adds a lot of wait and is a lot for PR testing :) You must wait X seconds till something happens or appears. For example in raiseHandToSpeak you click unmute, and wait 2000 seconds to check for the notification. You need to wait up to two seconds till the notification appears, there are plenty of examples in the code about that. Maybe like https://github.com/jitsi/jitsi-meet-torture/blob/527babf69b7b09a309a280684325ff727897b326/src/test/java/org/jitsi/meet/test/pageobjects/web/Notifications.java#L82
> 2. Please use separate methods for different tested cases, check the lobby test for example, how to order them.
> 3. * First test case, the happy one:
>      3 participants. Participant 1 enable av moderation. Participant 2 asks to unmute. Participant 1 sees the dialog, and allows it and checks that the notification disappeared (this is with participant pane closed).
>      Participant 3 asks to unmute, and moderator allowes it from within the participant pane (this way we test both UI components). In both cases make sure that the participant can unmute.
> 
> * Second test case:
>   You can hangup participant 2 and 3 and join them again and make sure they started muted. Ask to unmute from both, then hangup the moderator and join him again. The new moderator will be seeing only one notification to allow someone to unmute. The new moderator can be participant2 or participant3 not guranteed which one - your latest PR test failed because the other one was moderator, I think http://pr-reports.jitsi.org/jitsi-meet-PR-test/9929/ff-chrome-reports/html/.
> 
> So put case1 and case2 in different methods.

Second case is still in progress.

damencho (approved):

PR #405, opened 2021-08-23
feat: Adds test for grant moderator and AV moderation.

damencho:


damencho:
https://github.com/jitsi/jitsi-meet/pull/9785

hristoterezov (approved):

PR #410, opened 2021-09-10
fix: Directly select audio-only mode.

damencho:
This fix failing test DesktopSharingTest.testAudioOnlyAndNonDominantScreenShare, so if we go through low definition we set video-quality/preferredVideoQuality to 180 which is sent as maxHeight:180 to bridge and as the screenshare is with low bandwidth the bridge start sending it after 16 seconds.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/AudioOnlyTest.java
+++ b/src/test/java/org/jitsi/meet/test/AudioOnlyTest.java
@@ -172,8 +172,9 @@ public class AudioOnlyTest
             TestUtils.waitForCondition(
                 participantDriver,
                 2,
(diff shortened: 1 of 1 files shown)


jallamsetty1 (approved):

jallamsetty1 (dismissed):

damencho:
The iframe API failed with org.openqa.selenium.interactions.MoveTargetOutOfBoundsException: move target out of bounds

org jitsi meet test IFrameAPITest-web participant2

jallamsetty1 (approved):

damencho:
I'm trying it now in https://github.com/jitsi/jitsi-meet-torture/pull/416

damencho:
Will disable the test as even the other PR and 20 seconds and we do not get the video :(

PR #416, opened 2021-09-14
fix: Directly select audio-only mode.

damencho:
This fix failing test DesktopSharingTest.testAudioOnlyAndNonDominantScreenShare, so if we go through low definition we set video-quality/preferredVideoQuality to 180 which is sent as maxHeight:180 to bridge and as the screenshare is with low bandwidth the bridge start sending it after 16 seconds.


damencho:
Will disable the test as it still needs jvb fixes.

PR #423, opened 2021-09-21
fix: Make sure we leave the conference to clear the password.

damencho:


bgrozev (approved):

damencho:
Hum ... seems this is not needed ...

PR #441, opened 2021-10-21
fix: Updates raisedHand as timestamp(long) from boolean

horymury:


quitrk (dismissed):

quitrk:
just curious, why not use a map of values?

horymury:
> just curious, why not use a map of values?

Discussed offline, not applicable for this case. Renamed raisedHand var to raisedHandTimestamp.

saghul (approved):

PR #449, opened 2021-11-10
fix(lobby) Remove notification check on admit/ reject

robertpin:
Remove second participant notification check for action from participants pane. Check only that participant 3 is admitted / rejected


saghul (dismissed):
LGTM, but why are we removing that notification?

robertpin:
> LGTM, but why are we removing that notification?

We already have that on the admit/ reject from the notification, so it's not really needed on admit/ reject from participants pane. The fact that the participant is admitted/ rejected is enough to test the functionality.
It's being removed because it causes the test to be flaky and fail sometimes for no reason

robertpin:
Jenkins test this please

damencho (approved):

PR #452, opened 2021-11-16
fix(MuteTest): Wait for p2p ICE to be connected.

jallamsetty1:


src/test/java/org/jitsi/meet/test/MuteTest.java
```diff
@@ -247,13 +247,17 @@ private void muteParticipant1BeforeParticipant2JoinsAndScreenshare(String enable
...
-        participant2.waitToJoinMUC();
-        participant2.waitForIceConnected();
-        participant2.waitForSendReceiveData(true, false);
 
-        // Wait for the media to switch over to the p2p connection.
         TestUtils.waitMillis(2000);
 
+        // Wait for the media to switch over to the p2p connection in the p2p test.
+        if (enableP2p.equals("true")) {
+            MeetUtils.waitForP2PIceConnected(participant1.getDriver());
```
bgrozev, line 255: Did you want `participant2.getDriver()`  here? 

nit: The brace should be on a new line.

jallamsetty1: Fixed the brace. It shouldn't matter if we check for ICE connection establishment on p1 or p2 right ?

bgrozev: Well...I'm not sure, it's confusing.

If we can safely call `waitForSendReceiveData` before ICE has completed, why do we need to separately wait for ICE? Otherwise, it does make a difference if we check p1 or p2.

jallamsetty1: In the P2P case, we are waiting for the P2P connection to be established. That would mean that the media has switched over to the p2p connection and then only we proceed with the real test.

jallamsetty1: The p2p media session on both p1 and p2 gets created only when p2 joins the call so checking for the p2p connection state on either of them should work, right ?

src/test/java/org/jitsi/meet/test/MuteTest.java
```diff
@@ -246,29 +246,39 @@ private void muteParticipant1BeforeParticipant2JoinsAndScreenshare(String enable
...
-        // Stop desktop share and unmute video and check for video again.
-        participant1.getToolbar().clickDesktopSharingButton();
-
-        participant1.getToolbar().clickVideoMuteButton();
-        participant2.getFilmstrip().assertVideoMuteIcon(participant1, false);
-        MeetUIUtils.waitForRemoteVideo(participant2.getDriver(), participant1.getEndpointId(), true);
+        // Skip the test in p2p mode for Firefox since p2p is disabled in code for Firefox.
+        if (!participant1.getType().isFirefox() || enableP2p.equals("false"))
+        {
+            WebParticipant participant2 = joinSecondParticipant(url2);
```
damencho, line 253: We need here after joinSecondParticipant - `participant2.waitToJoinMUC();`,

damencho: But yeah ... if we are waiting for the p2p ice probably we can skip it ... if the check works with one participant ...

jallamsetty1: yes ideally we wouldn't need it since we are waiting for ice, pushed a commit for it.

damencho (approved):

damencho (approved):

PR #457, opened 2021-11-26
Revert "feat(config) replaced interfaceConfig with config on DisplayNameTest"

saghul:
Reverts jitsi/jitsi-meet-torture#456


saghul:
No need.

PR #465, opened 2021-12-10
feat(disableSelfView) Added tests for disable self view feature

robertpin:


src/test/java/org/jitsi/meet/test/DisableSelfViewTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+        ensureTwoParticipants();
+        participant1 = getParticipant1();
+        participant2 = getParticipant2();
+    }
+
+    @Test
+    public void testHideSelfViewFromMenu()
+    {
+        checkSelfViewDisplayed(false);
```
saghul, line 67: Can you change the name of this function? It reads as it it checks if it's displayed, but the parameter is negated (hidden) so it doesn't read well in English.

saghul:
Looks like this test fails on the Firefox run?

robertpin:
> Looks like this test fails on the Firefox run?

Fixed

saghul (approved):

PR #469, opened 2021-12-17
fix(moderatedRoomTest) Fix test

robertpin:


damencho (approved):

damencho:
This is not tested by the PR tests. Maybe we need to add it ... merging.

PR #482, opened 2022-01-21
fix: Temporary disable lobby tests.

damencho:


saghul (approved):

bgrozev:
I think we should have documentation on *why* tests are disabled. We have tests disabled in various places and we've lost track of why.

damencho:
@tudor-phd is working on it actively and will enable it once that is fixed.

PR #507, opened 2022-04-15
fix(stage-filmstrip) Fix active participant test

robertpin:


robertpin:
Jenkins test this please

robertpin:
Jenkins test this please

robertpin:
Jenkins test this please

robertpin:
Jenkins test this please

horymury (dismissed):

src/test/java/org/jitsi/meet/test/ActiveSpeakerTest.java
```diff
@@ -91,9 +91,18 @@ private void assertOneDominantSpeaker(WebDriver driver)
             driver.findElements(By.xpath(
                 "//span[contains(@class,'dominant-speaker')]"
             ));
+        List<WebElement> nonStage = new ArrayList<>();
+        for (WebElement el : dominantSpeakerIndicators)
```
hristoterezov, line 95: I'm wondering can we do the whole thing with XPath only?

Maybe we can have `id`s on the filmstrip containers and include in the XPath that there must be a parent with a specific `id` or maybe even use the current class of the stage filmstrip container in order to not include it in the XPath selection?

hristoterezov (changes requested):
Left a comment and suggestion.

src/test/java/org/jitsi/meet/test/ActiveSpeakerTest.java
robertpin: Yes, I think using xpath is a good idea. I'll update the code

robertpin:
Jenkins test this please

robertpin:
Jenkins test this please

robertpin:
Jenkins test this please

hristoterezov (approved):

PR #512, opened 2022-05-23
Remove button check which no longer apply

saghul:


damencho:
I will examine the PR tests before emerging this.

damencho (approved):

PR #521, opened 2022-07-19
Add flag to use Chrome "Black Frame Decoder", skipping video decode complexity.

JonathanLennox:
Enabled for Malleus, disabled for other tests.

### DIFF
--- a/scripts/malleus.sh
+++ b/scripts/malleus.sh
@@ -5,7 +5,7 @@ if [ -n "$DEBUG" ]; then
 fi
 
 usage() {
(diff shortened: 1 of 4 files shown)


scripts/malleus.sh
```diff
@@ -86,6 +86,10 @@ set_defaults() {
       MALLEUS_USE_STAGE_VIEW=false
     fi
 
+    if [ -z "$MALLEUS_ENABLE_DECODING"; then
```
bgrozev, line 89: Is this missing a `]`?

JonathanLennox:
Note this flag hasn't actually landed in Chrome yet, so don't merge this until it does.  See https://webrtc-review.googlesource.com/c/src/+/267820/14

src/test/java/org/jitsi/meet/test/web/WebParticipantFactory.java
```diff
@@ -254,6 +254,11 @@ else if (participantType == ParticipantType.edge)
                 ops.addArguments("window-size=1200x600");
             }
 
+            if (options.getDisableDecoding())
+            {
+                ops.addArguments("force-fieldtrials=WebRTC-BlackFrameDecoder/Enabled/");
```
fippo, line 259: beware, the current plan is to enable this as a separate flag for consistency with other decoders:
`--use-black-frame-decoder-for-peer-connection`
-- will update the webrtc CL accordingly

PR #528, opened 2022-08-23
fix(audio-only): Verify that participant is muted even after audio-on…

jallamsetty1:
…ly mode is disabled.

We do not automatically unmute the camera anymore when audio-only mode is disabled. If the user was video muted before audio-only was enabled, they will continue to stay video muted after audio-only mode is disabled.

### DIFF
--- a/src/test/java/org/jitsi/meet/test/AudioOnlyTest.java
+++ b/src/test/java/org/jitsi/meet/test/AudioOnlyTest.java
@@ -60,23 +60,9 @@ public class AudioOnlyTest
     }
 
     /**
-     * Verifies that participant1 cannot video unmute while in audio only mode.
-     */
-    @Test(dependsOnMethods = { "enableAudioOnlyAndCheck" })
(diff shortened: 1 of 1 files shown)


damencho (dismissed):

src/test/java/org/jitsi/meet/test/AudioOnlyTest.java
```diff
@@ -59,24 +59,10 @@ public void enableAudioOnlyAndCheck()
             true);
     }
 
-    /**
-     * Verifies that participant1 cannot video unmute while in audio only mode.
-     */
-    @Test(dependsOnMethods = { "enableAudioOnlyAndCheck" })
-    public void videoUnmuteDisabledInAudioOnly()
-    {
-        getParticipant1().getToolbar().clickAudioMuteButton();
```
jallamsetty1, line 68: The impl doesn't match the description. Also, user is allowed to unmute while in audio only mode.

damencho (approved):

PR #531, opened 2022-09-16
feat(multi-stream): Add desktop sharing testcases for multi-stream mode.

jallamsetty1:


src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
```diff
@@ -46,14 +56,42 @@ public void testDesktopSharingStart()
     {
         JitsiMeetUrl url = getJitsiMeetUrl()
             .appendConfig("config.filmstrip.disableStageFilmstrip=true")
-            .appendConfig("config.flags.sourceNameSignaling=false")
-            .appendConfig("config.flags.sendMultipleVideoStreams=false");
-        ensureOneParticipant(url);
+            .appendConfig("config.p2p.enabled=true");
 
+        ensureOneParticipant(url);
```
damencho, line 61: You should not need this if you have ensureTwoParticipants after that. Not that having it causes any issues :)

damencho (approved):
This PR failure we see it again
```
10:57:02 FAIL: Unexpected SRTP failures in node SrtcpDecryptNode.
10:57:02 FAIL: Too many discarded packets from node SrtcpDecryptNode: 1
```
@bgrozev @JonathanLennox if its expected can we bump the threshold :)

src/test/java/org/jitsi/meet/test/DesktopSharingTest.java
jallamsetty1: Hmm I don't get it. I am hanging up all the participants and then calling ensureOneParticipant to add the first participant. I am toggling screenshare on/off before adding the other two.

damencho: No, I mean ... look what ensureTwoParticipants do: https://github.com/jitsi/jitsi-meet-torture/blob/92b67127158b917d3a939bc9230666652fcf5252/src/test/java/org/jitsi/meet/test/web/WebTestBase.java#L151
It calls ensureOneParticipant :) What I mean is that there is no point of having that call of ensureOneParticipant :)

damencho (approved):

PR #546, opened 2022-11-02
fix(external_api) - fix tests breakoutRooms

dudumanbogdan:
- fix tests for https://github.com/jitsi/jitsi-meet/pull/12472


dudumanbogdan:
Closing as there is no need for a fix; the problem was with the new parameter set in lua

PR #547, opened 2022-11-03
feat: Adds lite mode param to malleus.

damencho:


JonathanLennox:
You should set `MALLEUS_USE_LITE_MODE` in malleus.sh's `set_defaults` function, too, or else if it's not passed you'll pass `-Dorg.jitsi.malleus.use_lite_mode=` (i.e., setting it to the empty string) to java, which might cause problems.

Is there any way that we could use lite mode only for receivers, thus removing the constraint that it can't be used with senders?

JonathanLennox (approved):

PR #565, opened 2023-02-07
Add "en" to accepted_languages

bugspencor:
This fixes torture tests for installations with non-english default languages.

See discussion https://community.jitsi.org/t/jitsi-meet-torture-and-non-english-default-language


bugspencor:
Cancel PR. I will create a new one from another repository.

PR #577, opened 2023-03-08
Enable media while in p2p and check if it is being received when the …

jallamsetty1:
…call switches to jvb.


jallamsetty1:
Jenkins test this please

damencho (approved):

PR #604, opened 2023-06-09
feat: Adds option to malleus to save participants logs.

damencho:
Not moving saveBrowserLogs to utils for now as it depends on outputLogsParentFolder that is handled in FailureListener.

### DIFF
--- a/scripts/malleus.sh
+++ b/scripts/malleus.sh
@@ -5,7 +5,7 @@ if [ -n "$DEBUG" ]; then
 fi
 
 usage() {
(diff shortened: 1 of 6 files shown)


scripts/malleus.sh
```diff
@@ -135,6 +135,7 @@ case $1 in
         --switch-speakers) if [ -n "$optvalue" ]; then MALLEUS_SWITCH_SPEAKERS=$optvalue; else MALLEUS_SWITCH_SPEAKERS=true; fi;;
         --use-stage-view) if [ -n "$optvalue" ]; then MALLEUS_USE_STAGE_VIEW=$optvalue; else MALLEUS_USE_STAGE_VIEW=true; fi;;
         --headless) if [ -n "$optvalue" ]; then MALLEUS_USE_HEADLESS=$optvalue; else MALLEUS_USE_HEADLESS=true; fi;;
+        --save-logs) if [ -n "$optvalue" ]; then MALLEUS_SET_SAVELOGS=$optvalue; else MALLEUS_SET_SAVELOGS=false; fi;;
```
JonathanLennox, line 138: This should be MALLEUS_SET_SAVELOGS=true so just passing `--save-logs` to the script causes it to save logs.

You'll also need to add MALLEUS_SET_SAVELOGS to the `set_defaults` function.

JonathanLennox (approved):

PR #606, opened 2023-06-21
feature: use default token in meet URL if provided

aaronkvanmeerten:


src/test/java/org/jitsi/meet/test/base/AbstractBaseTest.java
```diff
@@ -242,7 +242,7 @@ public JitsiMeetUrl getJitsiMeetUrl()
         String token = System.getProperty(TOKEN_PNAME);
 
         JitsiMeetUrl u = participants.getJitsiMeetUrl().setRoomName(currentRoomName);
-        if (token.length() > 0)
+        if (token != null)
         {
```
damencho, line 246: Or you want to check for both which is done with `StringUtils.isBlank` or `StringUtils.isNotBlank`

damencho, line 246: Or you want to check for both which is done with `StringUtils.isBlank` or `StringUtils.isNotBlank`

aaronkvanmeerten: Good point.

JonathanLennox (approved):

PR #612, opened 2023-08-22
feature: workflow for building grid images

aaronkvanmeerten:


sawall (approved):

damencho:
But this will push a new version every day?

aaronkvanmeerten:
> But this will push a new version every day?

No only when manually triggered

damencho (approved):

PR #616, opened 2023-08-24
fix: fix end meeting tests

quitrk:


damencho (dismissed):

damencho:
But please merge it at the same time with the related PR :)

PR #636, opened 2024-03-12
feat: Add a tc for p2p when both audio and video are muted at join time.

jallamsetty1:


src/test/java/org/jitsi/meet/test/StartMutedTest.java
```diff
@@ -261,4 +261,102 @@ public void startWithAudioMutedCanUnmute()
...
+            participant2.getDriver(),
+            participant1.getDriver(),
+            "participant1",
+            false);
+
+        // Unmute p2's audio and video and check on p1.
+        MeetUIUtils.unmuteVideoAndCheck(participant2, participant1);
+        participant2.getToolbar().clickAudioUnmuteButton();
+
+        participant1.getLargeVideo().isVideoPlaying();
```
bgrozev, line 320: Are you using this for a side effect or to assert something? The doc says it just returns a boolean

src/test/java/org/jitsi/meet/test/StartMutedTest.java
```diff
@@ -261,4 +261,102 @@ public void startWithAudioMutedCanUnmute()
...
+        participant2.getToolbar().clickVideoMuteButton();
+
+        participant3.hangUp();
+
+        participant1.getParticipantsPane().assertIsParticipantVideoMuted(participant2, true);
+        participant2.getToolbar().clickVideoMuteButton();
+
+        // Check if p2's video is playing on p1.
+        participant1.getParticipantsPane().assertIsParticipantVideoMuted(participant2, false);
+        participant1.getLargeVideo().isVideoPlaying();
```
bgrozev, line 360: Same as above

src/test/java/org/jitsi/meet/test/StartMutedTest.java
jallamsetty1: Oops, I just copy pasted from the existing testcase, will fix it.

bgrozev (approved):

bgrozev (dismissed):

PR #653, opened 2024-09-25
fix(getAccessibilityCSSSelector): Finding the matching aria-label

emrahcom:
This PR is related with jitsi/jitsi-meet#14997
It fixes the test to check the related PR correctly.

`Aria-label` is not static for the participant button after the update. So, looking for the whole match doesn't work. This PR allows to find the matching tag by checking the match at the beginning.


damencho:
Jenkins test this please.

damencho (approved):

PR #654, opened 2024-09-30
fix: Make the toolbar visible by using the new config parameter

emrahcom:
This PR is related with https://github.com/jitsi/jitsi-meet/pull/15054
It fixes the test tool to check the related PR correctly.

There is a new config parameter (`config.toolbarConfig.alwaysVisible`) to make the toolbar always visible during tests. The old one doesn't work anymore. So, the test tool cannot find some buttons in some conditions if the old one is used.


damencho (approved):

damencho:
Jenkins test this please.

